PR #5, opened 2025-08-08
config: update pint.json to use laravel pckage rules

pushpak1300:
This MR add changes to Pint config. 

The pint config has been copied form https://github.com/laravel/framework

### DIFF
--- a/pint.json
+++ b/pint.json
@@ -1,6 +1,171 @@
 {
-    "preset": "laravel",
-    "exclude": [
-        "stubs"
+    "preset": "empty",
+    "rules": {
+        "align_multiline_comment": true,
+        "array_indentation": true,
+        "array_syntax": {
+            "syntax": "short"
+        },
+        "binary_operator_spaces": {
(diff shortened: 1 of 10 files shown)


ashleyhindle:
One thing I’m not sure about with this is the CI because we used the
package template to create Boost initially, it uses other templates for
style checking in .GitHub/workflows

Will take a proper look shortly 👌




Ashley Hindle | Senior Software Engineer

***@***.***

www.laravel.com

pushpak1300:
CI Pipelines are failing mainly because of the  package issue. taking the look

ashleyhindle:
> CI Pipelines are failing mainly because of the package issue. taking the look

Looking at the workflow, it loads in style checking like this:

```
jobs:
  lint:
    uses: laravel/.github/.github/workflows/coding-standards.yml@main
```

And that just runs pint with no config options:

```
      - name: Run Pint
        run: pint
```

So it should be fine, if this pint config is better

PR #7, opened 2025-08-08
chore: improve composer dependencies.

pushpak1300:
We were supporting laravel 10 whch are EOL. 

Updated some basic depedecies.


ashleyhindle:
I purposely want to support Laravel 10 - it's EOL, but lots of people are still using it and we want to help the most people.

PR #9, opened 2025-08-10
refactor: cleanup BoostServiceProvider

pushpak1300:
This MR cleanup stuff in `BoostServiceProvider.php` .

- Cleanup some code
- Added comments to methods

### DIFF
--- a/src/BoostServiceProvider.php
+++ b/src/BoostServiceProvider.php
@@ -69,11 +69,6 @@ class BoostServiceProvider extends ServiceProvider
         $this->hookIntoResponses($router);
     }
 
-    /**
-     * Setup the resource publishing groups for Boost.
-     *
-     * @return void
-     */
     protected function registerPublishing(): void
     {
         if ($this->app->runningInConsole()) {
@@ -83,11 +78,6 @@ class BoostServiceProvider extends ServiceProvider
         }
     }
 
-    /**
-     * Register the Boost Artisan commands.
-     *
-     * @return void
-     */
     protected function registerCommands(): void
     {
         if ($this->app->runningInConsole()) {
@@ -99,11 +89,6 @@ class BoostServiceProvider extends ServiceProvider
         }
     }
 
-    /**
-     * Register the Boost routes.
-     *
-     * @return void
-     */
     private function registerRoutes(): void
     {
         Route::post('/_boost/browser-logs', function (Request $request) {
@@ -173,11 +158,6 @@ class BoostServiceProvider extends ServiceProvider
         ]);
     }
 
-    /**
-     * Register a Blade directive for the browser logger script.
-     *
-     * @param BladeCompiler $bladeCompiler
-     */
     protected function registerBladeDirectives(BladeCompiler $bladeCompiler): void
     {
         $bladeCompiler->directive('boostJs', fn () => '<?php echo \\Laravel\\Boost\\Services\\BrowserLogger::getScript(); ?>');


src/BoostServiceProvider.php
```diff
@@ -62,19 +65,29 @@ public function boot(Router $router): void
...
+     *
+     * @return void
+     */
     protected function registerPublishing(): void
     {
         if ($this->app->runningInConsole()) {
             $this->publishes([
                 __DIR__.'/../config/boost.php' => config_path('boost.php'),
-            ], 'boost-config');
+            ], ['boost-config', 'config']);
```
ashleyhindle: What does this change do?

src/BoostServiceProvider.php
```diff
@@ -62,19 +65,29 @@ public function boot(Router $router): void
...
         }
     }
 
+    /**
```
ashleyhindle: These kind of comments aren't useful though right? We know the return type and what the method does because of the name and types.

Is this a format we should be following? Same for all comments here, what are your thoughts on what's actually sensible and easy to read, separate to what the framework does?

pushpak1300: Ideally I don't like to repeat things.  I prefer only method with great naming and proper typing. I adde this to follow overall framework/packages styles and be consitent with the structure. I'm happy to remove it though.

src/BoostServiceProvider.php
pushpak1300: Not mandatory to check, but a nice-to-have.

Users can publish files using either of these tags:
	•	`php artisan vendor:publish --tag=boost-config`
	•	`php artisan vendor:publish --tag=config`

This gives them flexibility in how they prefer to organize their publishing commands.

ashleyhindle: `config` feels like it will be a bit ambiguous to users, many other packages are registering things to publish with different tags right? I'm not sure we can claim to offer the 'config' to be published, we only offer the Boost config

src/BoostServiceProvider.php
ashleyhindle: What do you think is most important in this repo? Should we follow the framework completely, or do what's most modern?

src/BoostServiceProvider.php
```diff
@@ -86,17 +99,29 @@ protected function registerCommands(): void
...
     {
         Route::post('/_boost/browser-logs', function (Request $request) {
             $logs = $request->input('logs', []);
             /** @var \Illuminate\Log\Logger $logger */
             $logger = Log::channel('browser');
 
-            /** @var array{type: 'error'|'warn'|'info'|'log'|'table'|'window_error'|'uncaught_error'|'unhandled_rejection', timestamp: string, data: array<mixed>, url:string, userAgent:string} $log */
+            /**
+             *  @var array{
+             *      type: 'error'|'warn'|'info'|'log'|'table'|'window_error'|'uncaught_error'|'unhandled_rejection',
```
ashleyhindle, line 101: Do you know if there's a nicer way to handle this? I've never liked it, but I'm not a phpdoc pro

I _guess_ we could add an Enum, but that _feels_ overkill

src/BoostServiceProvider.php
pushpak1300: Okay. As It's option I can revert this back. Thanks

src/BoostServiceProvider.php
pushpak1300: Tricky question, haha. I’d say since we’ve been following a few modern practices in this package like strict typing we should continue following modern concepts and remove redundant comments. Most older packages are hard to refactor into a modern structure, so whenever we start something new, it makes sense to adopt the updated structure from the beginning. That's my opinion happy to follow what taylor and you thinks here.

ashleyhindle: 😁

I agree with that, let's do that 👍

src/BoostServiceProvider.php
pushpak1300: Agree this is correct typing only. 

As this is the single usage using Enum is definitely overkill.

PR #10, opened 2025-08-10
config: add project purpose and browser logs watcher options to boost.php and organnise config in laravel style

pushpak1300:
This MR add support for the cleanin up the config for Boost Package. COnfig is now arragned in Laravel way with some better naming

### DIFF
--- a/src/BoostServiceProvider.php
+++ b/src/BoostServiceProvider.php
@@ -142,7 +142,7 @@ class BoostServiceProvider extends ServiceProvider
      * "data":[
      * {"message":"Unhandled Promise Rejection","reason":{"name":"TypeError","message":"NetworkError when attempting to fetch resource.","stack":""}}]
      *
-     * @param  array<mixed>  $data
+     * @param array<mixed> $data
      */
     protected function buildLogMessageFromData(array $data): string
     {
@@ -176,7 +176,7 @@ class BoostServiceProvider extends ServiceProvider
     /**
      * Register a Blade directive for the browser logger script.
      *
-     * @param  BladeCompiler  $bladeCompiler
+     * @param BladeCompiler $bladeCompiler
      */
     protected function registerBladeDirectives(BladeCompiler $bladeCompiler): void
     {
--- a/pint.json
+++ b/pint.json
@@ -111,7 +111,7 @@
         "phpdoc_align": {
             "align": "left",
             "spacing": {
-                "param": 2
+                "param": 1
             }
         },
         "phpdoc_indent": true,
--- a/src/Concerns/MakesHttpRequests.php
+++ b/src/Concerns/MakesHttpRequests.php
@@ -30,7 +30,7 @@ trait MakesHttpRequests
     }
 
     /**
-     * @param  array<string, mixed>  $json
+     * @param array<string, mixed> $json
      */
     public function json(string $url, array $json): Response
     {
--- a/src/Contracts/Ide.php
+++ b/src/Contracts/Ide.php
@@ -10,8 +10,8 @@ interface Ide
     // Things to note: supports relative (absolute path required)? global mcp only? Prefer local file, but if global only we have to add the project name to the server name
 
     /**
-     * @param  array<int, string>  $args
-     * @param  array<string, string>  $env
+     * @param array<int, string> $args
+     * @param array<string, string> $env
      */
     public function installMcp(string $key, string $command, array $args = [], array $env = []): bool;
 }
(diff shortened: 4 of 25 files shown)


config/boost.php
```diff
@@ -3,6 +3,28 @@
...
+    | Boost Project Purpose
+    |--------------------------------------------------------------------------
+    |
+    | This is the purpose of the Boost project helping AI understand context
+    | the project and its purpose and be guidelines more specific
+    | project. This should be set to a very specific.
+    |
+    */
+
     'project_purpose' => null,
```
ashleyhindle: I've disabled project purpose in the install command, so we don't need to worry about this right now, but we'll need to triple check adding the project purpose to the config in `InstallCommand` still works with this format in future

src/BoostServiceProvider.php
```diff
@@ -174,13 +174,13 @@ protected function registerBrowserLogger(): void
     }
 
     /**
-     * Register a Blade directive for the browser logger script
+     * Register a Blade directive for the browser logger script.
      *
-     * @param BladeCompiler $bladeCompiler
+     * @param  BladeCompiler  $bladeCompiler
```
ashleyhindle: Are double spaces a thing?

config/boost.php
```diff
@@ -3,6 +3,28 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Boost Browser Logs Watcher Switch
+    |--------------------------------------------------------------------------
+    |
+    | The following option may be used to enable or disable browser logs watcher
+    | functionality which simply provides a single and convenient way to
+    | enable or disable browser logs functionality in Boost.
+    */
+    'boost_browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
```
ashleyhindle: We're already in 'boost', this shouldn't have boost in it

src/BoostServiceProvider.php
```diff
@@ -198,7 +198,7 @@ private function mapJsTypeToPsr3Level(string $type): string
 
     private function hookIntoResponses(Router $router): void
     {
-        if (config('boost.browser_logs', true) === false) {
+        if (! config('boost.boost_browser_logs_watcher', true)) {
```
ashleyhindle: Same comment re: config - we're already in the 'boost' context, so the key shouldn't have boost in it

pushpak1300: Agree. My Bad. Reverting to `browser_logs_watcher`

src/BoostServiceProvider.php
pushpak1300: This has been enforced by Pint based on pint config. Reverting those changes in https://github.com/laravel/boost/pull/11

config/boost.php
pushpak1300: Fixed.

config/boost.php
pushpak1300: okay then reverting this now. I have removed this for now.

src/BoostServiceProvider.php
ashleyhindle: Oh wow, very odd! Merged in #11 so should be good 🙏

config/boost.php
```diff
@@ -3,6 +3,16 @@
...
+    |--------------------------------------------------------------------------
+    | Boost Browser Logs Watcher Switch
+    |--------------------------------------------------------------------------
+    |
+    | The following option may be used to enable or disable browser logs watcher
+    | functionality which simply provides a single and convenient way to
+    | enable or disable browser logs functionality in Boost.
+    */
+
+    'browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
```
ashleyhindle, line 17: I wonder if this would be better named as `browser_log_streamer` or `stream_browser_logs`

'Watcher' makes me think of tools that poll every second or so

pushpak1300: The naming has been inspired by the telescope package where it's also does the similar stuff (watching the debug logs)

ashleyhindle: Let's stick with it for now then 👌

pushpak1300: But `browser_log_streamer` or `browser_log_streamer`  are little confusing  to me. Even though we are streaming browser logs to application the end usage(feature) is about enabling the browser logging so we can view the logs.

ashleyhindle:
Partially. It’s also about whether to inject the middleware that enables
the streaming or not. We don’t inject if it’s false right, so
“stream_browser_logs” is valid.

But happy with the current key and merged anywho 👌




Ashley Hindle | Senior Software Engineer

***@***.***

www.laravel.com

PR #12, opened 2025-08-11
Cleanup Install Command 1

pushpak1300:
Cleanup some basic structrue of install command.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -183,9 +183,9 @@ class InstallCommand extends Command
         $allData = array_merge($ideNames, $agentNames, $boostFeatures, $guidelines);
 
         // Create compact CSV string and base64 encode
-        $trackingData = base64_encode(implode(',', $allData));
+        $installData = base64_encode(implode(',', $allData));
 
-        $link = $this->hyperlink($label, 'https://boost.laravel.com/installed/?d='.$trackingData);
+        $link = $this->hyperlink($label, 'https://boost.laravel.com/installed/?d='.$installData);
 
         $text = 'Enjoy the boost 🚀 ';
         $paddingLength = (int) (floor(($this->terminal->cols() - mb_strlen($text.$label)) / 2)) - 2;


src/Console/InstallCommand.php
```diff
@@ -23,85 +23,103 @@
...
-    protected array $detectedProjectIdes = [];
+    private Terminal $terminal;
 
-    protected bool $enforceTests = true;
+    /** @var Collection<int, \Laravel\Boost\Contracts\Agent> */
+    private Collection $agentsToInstallTo = collect();
 
     /** @var Collection<int, \Laravel\Boost\Contracts\Ide> */
-    protected Collection $idesToInstallTo;
+    private Collection $idesToInstallTo = collect();
```
ashleyhindle: I don't think you can do this in PHP 8.1, having a default value be a function, we'll need to initialize this in construct

src/Console/InstallCommand.php
```diff
@@ -23,85 +23,103 @@
...
 
-    protected ApplicationDetector $appDetector;
+    private array $detectedProjectAgents = [];
 
     private string $greenTick;
 
     private string $redCross;
 
-    private Terminal $terminal;
+    public function handle(ApplicationDetector $appDetector, Herd $herd, Roster $roster, Terminal $terminal): void
```
ashleyhindle: This method is so clean now, love it 🧼

src/Console/InstallCommand.php
ashleyhindle: Or we _can_ use `new Collection()`, that's supported in 8.1 I believe

src/Console/InstallCommand.php
```diff
@@ -23,85 +23,103 @@
...
+    }
 
-        $this->intro();
-        $this->detect();
-        $this->query();
-        $this->enact();
-        $this->outro();
+    private function displayBoostHeader(): void
+    {
+        note($this->boostLogo());
```
ashleyhindle, line 94: Do we need `note` on the logo?

pushpak1300: can use `$this->line` but Note is natively supported by laravel prompt so want to be consistent with it while we display anything.

src/Console/InstallCommand.php
pushpak1300: Agree. Setting IDE on 8.1 PHP 🥲

PR #21, opened 2025-08-12
Refactor Tools to improve consistency and readability

pushpak1300:
This pull request refactors the code across tools to enhance consistency and readability. No existing features or behaviors are altered. The changes aim to improve maintainability and consistency

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema/PostgreSQLSchemaDriver.php
+++ b/src/Mcp/Tools/DatabaseSchema/PostgreSQLSchemaDriver.php
@@ -7,7 +7,7 @@ namespace Laravel\Boost\Mcp\Tools\DatabaseSchema;
 use Exception;
 use Illuminate\Support\Facades\DB;
 
-class PostgresSQLSchemaDriver extends DatabaseSchemaDriver
+class PostgreSQLSchemaDriver extends DatabaseSchemaDriver
 {
     public function getViews(): array
     {
--- a/src/Mcp/Tools/DatabaseSchema/SchemaDriverFactory.php
+++ b/src/Mcp/Tools/DatabaseSchema/SchemaDriverFactory.php
@@ -14,7 +14,7 @@ class SchemaDriverFactory
 
         return match ($driverName) {
             'mysql', 'mariadb' => new MySQLSchemaDriver($connection),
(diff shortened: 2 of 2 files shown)


src/Mcp/Methods/CallToolWithExecutor.php
```diff
@@ -4,45 +4,43 @@
...
 use Laravel\Mcp\Server\Transport\JsonRpcRequest;
 use Laravel\Mcp\Server\Transport\JsonRpcResponse;
+use Throwable;
 
 class CallToolWithExecutor implements Method
 {
     /**
      * Handle the JSON-RPC tool/call request with process isolation.
      *
-     * @return JsonRpcResponse|Generator<JsonRpcNotification|JsonRpcResponse>
```
ashleyhindle, line 22: Is your phpstan happy with this change? I remember fighting static analysis here 😆

src/Mcp/Tools/DatabaseConnections.php
```diff
@@ -19,7 +19,6 @@ public function description(): string
 
     public function schema(ToolInputSchema $schema): ToolInputSchema
```
ashleyhindle, line 20: Is this the default? Wonder if we can delete this method, or if we should update the MCP server package so it is the default?

src/Mcp/Methods/CallToolWithExecutor.php
pushpak1300: I didn't got the error with this though.  Which level are you using ?

ashleyhindle: Oh perfect. Whatever is in the phpstan config. But I did bump it one previously and fix some things, so it may be at the next level, or I'm misremembering 😆

src/Mcp/Tools/DatabaseConnections.php
ashleyhindle: Ah it isn't https://github.com/laravel/mcp/blob/main/src/Server/Tool.php#L33

But it should be, so I'm going to change that in the MCP package so we can remove these empty input schemas

ashleyhindle: Done - https://github.com/laravel/mcp/commit/9d9c7514c973497b7e003685b4e9792149b1480a

PR #25, opened 2025-08-12
Refactor SearchDocs tests to use Http facade

pushpak1300:
### Description

This pull request refactors the `SearchDocs` tests by replacing `Mockery` with Laravel's `Http` facade for a cleaner and more maintainable approach. Key changes include:

- Replaced `Mockery` with `Http::fake` and `Http::assertSent`.
- Updated `MakesHttpRequests` to support the `testing` environment with proper SSL verification handling.
- Added `guzzlehttp/guzzle` as a dependency in `composer.json`.

These updates simplify the test logic and modernize the approach for handling HTTP requests in tests.

### DIFF
--- a/src/BoostServiceProvider.php
+++ b/src/BoostServiceProvider.php
@@ -126,9 +126,6 @@ class BoostServiceProvider extends ServiceProvider
     /**
      * Build a string message for the log based on various input types. Single-dimensional, and multi:
      * "data": {"message":"Unhandled Promise Rejection","reason":{"name":"TypeError","message":"NetworkError when attempting to fetch resource.","stack":""}}]
-     *
-     * @param array $data
-     * @return string
      */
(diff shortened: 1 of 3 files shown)


ashleyhindle:
Is guzzle a dependency? I thought Http found any compatible PSR-7 client? 

Trying to keep the boost dependencies _really_ small, so want to be 100% certain here

composer.json
```diff
@@ -20,6 +20,7 @@
     ],
     "require": {
         "php": "^8.1|^8.2",
+        "guzzlehttp/guzzle": "^7.9",
```
ashleyhindle, line 23: Seems like this is needed as 7.8.2 is required in laravel/framework/illuminate/http

I'll check which version is supported in php 8.1 to see how workable this is

PR #33, opened 2025-08-12
Use absolute path for PHPStorm and relative paths for others

ashleyhindle:
- Linear: https://linear.app/laravel/issue/AI-68/how-can-we-work-around-requiring-absolute-paths-needed-for-launch

# What & Why
Developers can't commit the mcp files to git if they have absolute paths in, as each developer has a unique setup.

All IDEs seem to support relative paths, except PHPStorm. This PR updates our MCP registration so we use 'absolute' for PHPStorm/Junie, but 'relative' for every other IDe.

# How I tested
  1. `boost:install` and chose Cursor/PHPStorm
  2. Checked the files

```
❯ grep artisan .junie/mcp/mcp.json .cursor/mcp.json
.junie/mcp/mcp.json:                "/Users/ashleyhindle/Code/scratch/flux-volt-folio-playground/artisan",
.cursor/mcp.json:                "./artisan",
```

# Note
- Removed Zed also as it was getting messy and we randomly added MCP/guidelines to it, even though it never showed up in the options.

### DIFF
--- a/src/Install/CodeEnvironment/CodeEnvironment.php
+++ b/src/Install/CodeEnvironment/CodeEnvironment.php
@@ -17,7 +17,9 @@ abstract class CodeEnvironment
 {
     public bool $useAbsolutePathForMcp = false;
 
-    public function __construct(protected readonly DetectionStrategyFactory $strategyFactory) {}
+    public function __construct(protected readonly DetectionStrategyFactory $strategyFactory)
+    {
+    }
 
     abstract public function name(): string;
 


src/Install/CodeEnvironment/CodeEnvironment.php
```diff
@@ -15,6 +15,8 @@
 
 abstract class CodeEnvironment
 {
+    public bool $absolute = false;
+
```
pushpak1300: Can we rename this to something more descriptive? 
Such as `requireAbsolutePath`  or `configureAbsolutePath` ?

src/Console/InstallCommand.php
```diff
@@ -479,7 +470,9 @@ private function installMcpServerConfig(): void
 
             if ($this->shouldInstallMcp()) {
                 try {
-                    $result = $mcpClient->installMcp('laravel-boost', 'php', ['./artisan', 'boost:mcp']);
+                    /** @phpstan-ignore property.notFound */
```
pushpak1300: /** @var CodeEnvironment $mcpClient */ instead of ignoring phpstan error we can define the instance of code environment.

ashleyhindle:
Thank you! Pushed changes for those @pushpak1300 🙌

PR #43, opened 2025-08-13
Update Pint Guideline to Use `--dirty` Flag

yitzwillroth:
Updates `.ai/pint/core.blade.php` to prefer usage of Pint with the `--dirty` flag for cleaner diffs/PRs.


ashleyhindle:
This makes sense to me, I like it! 👌

PR #52, opened 2025-08-13
Support `false` as the content argument

arondeparon:
This previously broke for responses that have the (valid!) `false` value content attribute value, which happens for `StreamedResponse` objects, for example.


ashleyhindle:
I just modified InjectBoost so this is now out of date, but I'll add this check to another PR 👌 Thanks for raising!

PR #61, opened 2025-08-13
Fix Herd detection

mpociot:
There is an issue in the current Herd detection as it's checking for the wrong files/paths on Mac and Windows.


ashleyhindle:
Thank you, and that's frustrating!

Issue #36, opened 2025-08-13
PhpStorm: Unable to activate Boost


jobjen02:
### Laravel Package Version

1.0.1

### Laravel Version

12.20

### PHP Version

8.4.0

### Database Driver & Version

Postgres 17

### Description

When I try to activate laravel-boost in my "MCP Settings" in PhpStorm, I get an red exclamation mark with the message: 

```
StandaloneCoroutine was cancelled

Could not open input file: /var/www/html/artisan
```

I'm using docker.

<img width="994" height="725" alt="Image" src="https://github.com/user-attachments/assets/f2022bae-3680-4537-99eb-9273e0384673" />

### Steps To Reproduce

Setup an laravel project with docker and install boost (I guess). Not sure why it's not working on our side. We don't have any special things within our Laravel project.

ashleyhindle:
Can you share the contents of your `.junie/mcp/mcp.json` file?

Boost should be ran from the host ideally, was boost:install ran within the docker container? 👌

jobjen02:
> Can you share the contents of your `.junie/mcp/mcp.json` file?
> 
> Boost should be ran from the host ideally, was boost:install ran within the docker container? 👌

Yes I did run it from the docker container. 

Here is my `.junie/mcp/mcp.json` file:
```
{
    "mcpServers": {
        "laravel-boost": {
            "command": "php",
            "args": [
                "/var/www/html/artisan",
                "boost:mcp"
            ]
        }
    }
}
```

ashleyhindle:
Ah ok, you might need to wrap that in a docker run command which _should_ work. Something like this:

```json
{
    "mcpServers": {
        "laravel-boost": {
            "command": "docker",
            "args": [
                "exec",
                "-i",
                "<container_name>",
                "php",
                "/var/www/html/",
                "artisan",
                "boost:mcp"
            ]
        }
    }
}
```

jobjen02:
Now I only get this error: `StandaloneCoroutine was cancelled`. This is my mcp.json:

```
{
    "mcpServers": {
        "laravel-boost": {
            "command": "docker",
            "args": [
                "exec",
                "-i",
                "shoplinkr-api-1",
                "php",
                "/var/www/html/",
                "artisan",
                "boost:mcp"
            ]
        }
    }
}
```

pratikkuikel:
Non docker environment.
can confirm that ./artisan in the terminal works, wasn't working before though. Had to chmod +x artisan.
And
[...]

PR #65, opened 2025-08-13
docs: add docker mcpServer example

calebdw:
Hello!

This adds an example for when using boost inside of a docker container.

From: https://github.com/laravel/boost/issues/36#issuecomment-3184361793

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,6 @@ Laravel Boost includes AI guidelines for the following packages and frameworks. 
 | Laravel Folio | core |
 | Enforce Tests | conditional |
 
-
 ## Available Documentation
 
 | Package | Versions Supported |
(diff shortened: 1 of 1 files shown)


calebdw:
Note, it would be better to use `sail` so you don't have to hardcode container names and artisan paths, but one person said that didn't work for them: https://github.com/laravel/boost/issues/36#issuecomment-3184557766

I would be curious to know why

CC: @riyuk

calebdw:
Actually, sail does appear to be working via this comment (and from my own brief testing): https://github.com/laravel/boost/issues/36#issuecomment-3185244581

pushpak1300:
@ashleyhindle I have tested this locally with sail project and it works the only thing we have to decide is do we want to support it directly via command https://github.com/laravel/boost/pull/83 or add it in readme, or both ?

jeffc-dev:
Not everybody uses Sail. A plain Docker example would be good to have as well.

pushpak1300:
Thanks for your pull request! 

PR #68, opened 2025-08-13
fix: use app name as project name instead of directory name

calebdw:
Hello!

When running inside of Docker, boost thinks the project name is `html` which is a little strange:

image


ashleyhindle:
Ah good point 🤔 Though when running a brand new project the app name is 'Laravel', which actually works in that sentence to be fair, but it's also inaccurate 😁

calebdw:
I guess we could check if the name is `Laravel` and fall back to the directory name?

ashleyhindle:
> I guess we could check if the name is `Laravel` and fall back to the directory name?

Aye I was just wondering about that, but I couldn't convince myself it was the right thing. `config('app.name');` is better for now and we can revisit this later if needed 💪

PR #83, opened 2025-08-13
fix: detect if project uses sail

calebdw:
Hello!

This detects if the project uses `sail` and automatically updates the mcp commands

Thanks!


nickpoulos:
I feel like we also need to update some other stubs or generated guidelines. 

Ex. The guidelines being generated within CLAUDE.md - these examples would fail without appending "sail" in front

```
### Vite Error
- If you receive an "Illuminate\Foundation\ViteException: Unable to locate file in Vite manifest" error, you can run `npm run build` or ask the user to run `npm run dev` or `composer run dev`.
```

calebdw:
Sure, but that's a slightly separate concern. The main thing is getting the mcp server to be utilized in the first place

Liamso:
+1 - surprised this wasn't picked up during dev. Thanks for the PR. 

Might be best to allow an override in the user's .env - some projects will have sail installed but unused.

gigerIT:
Can we also add Sail support for the guidelines files in this PR?
Currently, it generates them with:

`php artisan test` 

Example:
```md
=== tests rules ===

## Test Enforcement

- Every change must be programmatically tested. Write a new test or update an existing test, then run the affected tests to make sure they pass.
- Run the minimum number of tests needed to ensure code quality and speed. Use `php artisan test` with a specific filename or filter.
</laravel-boost-guidelines>
```

Instead of:
`sail test`

nikspyratos:
I think Sail docs can be split out into another PR, but would it not be better to instruct the agent to just use `sail artisan` for _all_ artisan commands instead of individual instructions in this scenario?

gigerIT:
> I think Sail docs can be split out into another PR, but would it not be better to instruct the agent to just use `sail artisan` for _all_ artisan commands instead of individual instructions in this scenario?

Agreed, then it could just be a variable, either `php` or `sail`, used as the base for all generated commands in the docs.

pushpak1300:
Should we add configuration here to support Laravel Sail?
I have several projects where I've installed Sail, but I end up running composer run dev locally instead. This is a very opinionated choice.
If we add Sail support. we can add it to installation options, we'd need to ask users whether they're using Sail or not. Simply detecting the package installation and auto-configuring based on that could lead to incorrect installations.

Curios to know your thoughts.

pushpak1300:
Thanks or your pull request ! 

Closing this in favour of #303

PR #93, opened 2025-08-14
fix: Prevent install command from breaking when `/tests` doesn't exist

### LINKED ISSUE #45: When the `tests` directory does not exist, `boost:install` fails
### Laravel Package Version

1.0.1

### Laravel Version

12.23.1

### PHP Version

8.4.11

### Database Driver & Version

_No response_

### Description

We use `nwidart/laravel-modules` so all of our tests live under modules/{module_name}/Tests and we do not have a top level `tests` directory. The `boost:install` command errors with the following:

```
php artisan boost:install

 ██████╗   ██████╗   ██████╗  ███████╗ ████████╗
 ██╔══██╗ ██╔═══██╗ ██╔═══██╗ ██╔════╝ ╚══██╔══╝
 ██████╔╝ ██║   ██║ ██║   ██║ ███████╗    ██║
 ██╔══██╗ ██║   ██║ ██║   ██║ ╚════██║    ██║
 ██████╔╝ ╚██████╔╝ ╚██████╔╝ ███████║    ██║
 ╚═════╝   ╚═════╝   ╚═════╝  ╚══════╝    ╚═╝

  ✦ Laravel Boost :: Install :: We Must Ship ✦ 

 Let's give propper a Boost


   Symfony\Component\Finder\Exception\DirectoryNotFoundException 

  The "{redacted_project_dir}}/tests" directory does not exist.

  at vendor/symfony/finder/Finder.php:646
    642▕             } elseif ($glob = glob($dir, (\defined('GLOB_BRACE') ? \GLOB_BRACE : 0) | \GLOB_ONLYDIR | \GLOB_NOSORT)) {
    643▕                 sort($glob);
    644▕                 $resolvedDirs[] = array_map($this->normalizeDir(...), $glob);
    645▕             } else {
  ➜ 646▕                 throw new DirectoryNotFoundException(\sprintf('The "%s" directory does not exist.', $dir));
    647▕             }
    648▕         }
    649▕ 
    650▕         $this->dirs = array_merge($this->dirs, ...$resolvedDirs);

      +15 vendor frames 

  16  artisan:35
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

Obviously we can re-add the folder with a `.gitkeep` but we do have a substantial test suite, just not in the default location. Can there be an option to provide a glob pattern for test directories so that the generated guidelines still enforce tests?

### Steps To Reproduce

To reproduce, run `boost:install` when there is no `tests` directory present in the top level of the laravel project.

sagalbot:
Currently, on `v1.0.9`, `php artisan boost:install` will fail for any project that doesn't contain a `tests` directory at the root of the project.

```
~/Sites/central install-laravel-boost* 20s
❯ php artisan boost:install

 ██████╗   ██████╗   ██████╗  ███████╗ ████████╗
 ██╔══██╗ ██╔═══██╗ ██╔═══██╗ ██╔════╝ ╚══██╔══╝
 ██████╔╝ ██║   ██║ ██║   ██║ ███████╗    ██║
 ██╔══██╗ ██║   ██║ ██║   ██║ ╚════██║    ██║
 ██████╔╝ ╚██████╔╝ ╚██████╔╝ ███████║    ██║
 ╚═════╝   ╚═════╝   ╚═════╝  ╚══════╝    ╚═╝

  ✦ Laravel Boost :: Install :: We Must Ship ✦ 

 Let's give Redacted Project Name a Boost

Symfony\Component\Finder\Exception\DirectoryNotFoundException 

The "/Users/sagalbot/Sites/central/tests" directory does not exist.

at vendor/symfony/finder/Finder.php:646
  642▕             } elseif ($glob = glob($dir, (\defined('GLOB_BRACE') ? \GLOB_BRACE : 0) | \GLOB_ONLYDIR | \GLOB_NOSORT)) {
  643▕                 sort($glob);
  644▕                 $resolvedDirs[] = array_map($this->normalizeDir(...), $glob);
  645▕             } else {
➜ 646▕                 throw new DirectoryNotFoundException(\sprintf('The "%s" directory does not exist.', $dir));
  647▕             }
  648▕         }
  649▕ 
  650▕         $this->dirs = array_merge($this->dirs, ...$resolvedDirs);

    +15 vendor frames 

16  artisan:35
    Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

The installer looks for tests within the directory to determine if the user should be prompted about AI writing tests. The strategy here makes sense - remove an install step if no tests are found, but the current implementation will fail for many projects.

I think the simplest possible solution here is to just ask the user. Yes, it's one extra step, but it puts control in the hands of the user. If you're installing into a fresh Laravel project, the arbitrary 6 test threshold means you won't be asked about AI test guidelines.

If you do want to use the presence of tests as a way to determine if this question should be asked, you could call `phpunit --list-tests` and check if the output is >=6 lines. That should work for jest and phpunit regardless of test location.

## Changes
- removes check for `/tests` (fixes install failure)
- changes flow so asking is the last step

Resolves #45

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -59,7 +59,7 @@ class InstallCommand extends Command
 
     private bool $enforceTests = true;
 
-    const MIN_TEST_COUNT = 2;
+    const MIN_TEST_COUNT = 6;
 
     private string $greenTick;
 
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -23,6 +23,7 @@ use Laravel\Prompts\Concerns\Colors;
 use Laravel\Prompts\Terminal;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Process\Process;
 
 use function Laravel\Prompts\intro;
 use function Laravel\Prompts\multiselect;
@@ -57,6 +58,8 @@ class InstallCommand extends Command
     private array $projectInstalledCodeEnvironments = [];
 
     private bool $enforceTests = true;
+
+    const MIN_TEST_COUNT = 2;
 
     private string $greenTick;
 
@@ -120,7 +123,7 @@ class InstallCommand extends Command
         $this->selectedBoostFeatures = $this->selectBoostFeatures();
         $this->selectedTargetMcpClient = $this->selectTargetMcpClients();
         $this->selectedTargetAgents = $this->selectTargetAgents();
-        $this->enforceTests = $this->determineTestEnforcement();
+        $this->enforceTests = $this->determineTestEnforcement(ask: false);
     }
 
     private function performInstallation(): void
@@ -196,13 +199,29 @@ class InstallCommand extends Command
      * won't have the CI setup to make use of them anyway, so we're just wasting their
      * tokens/money by enforcing them.
      */
-    protected function determineTestEnforcement(): bool
+    protected function determineTestEnforcement(bool $ask = true): bool
     {
-        return select(
-            label: 'Should AI always create tests?',
-            options: ['Yes', 'No'],
-            default: 'Yes'
-        ) === 'Yes';
+        $hasMinimumTests = false;
+
+        if (file_exists(base_path('vendor/bin/phpunit'))) {
+            $process = new Process(['php', 'artisan', 'test', '--list-tests']);
+            $process->run();
+
+            /** Count the number of tests - they'll always have :: between the filename and test name */
+            $hasMinimumTests = collect(explode("\n", trim($process->getOutput())))
+                ->filter(fn ($line) => str_contains($line, '::'))
+                ->count() >= self::MIN_TEST_COUNT;
+        }
+
+        if (! $hasMinimumTests && $ask) {
+            $hasMinimumTests = select(
+                label: 'Should AI always create tests?',
+                options: ['Yes', 'No'],
+                default: 'Yes'
+            ) === 'Yes';
+        }
+
+        return $hasMinimumTests;
     }
 
     /**
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -204,7 +204,7 @@ class InstallCommand extends Command
         $hasMinimumTests = false;
 
         if (file_exists(base_path('vendor/bin/phpunit'))) {
-            $process = new Process(['php', 'artisan', 'test', '--list-tests']);
+            $process = new Process([PHP_BINARY, 'artisan', 'test', '--list-tests'], base_path());
             $process->run();
 
             /** Count the number of tests - they'll always have :: between the filename and test name */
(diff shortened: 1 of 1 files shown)


pushpak1300:
@sagalbot 
Initially, we're aiming to keep the number of questions during installation to a minimum, which is why we have the ask option available but haven't enabled it yet.
I was thinking if we could make the failure avoid when the test directory doesn't exist. For example, we could implement something like this:
```php
$testsPath = base_path('tests');
$hasMinimumTests = false;
if (is_dir($testsPath)) {
    $hasMinimumTests = Finder::create()
        ->in($testsPath)
        ->files()
        ->name('*.php')
        ->count() > 6;
}
```
What do you think about this suggestion?

ashleyhindle:
This would be my preference too for right now, checking if `tests` exists before searching.

We added the `ask` option because we _do_ want to ask at some point when needed, but it's a bit in the way at the minute and we need to find a better DX for that & other questions 👌

src/Console/InstallCommand.php
```diff
@@ -198,11 +201,17 @@ private function hyperlink(string $label, string $url): string
...
+
+        if (file_exists(base_path('vendor/bin/phpunit'))) {
+            $process = new Process(['php', 'artisan', 'test', '--list-tests']);
+            $process->run();
+
+            /** Count the number of tests - they'll always have :: between the filename and test name */
+            $hasMinimumTests = collect(explode("\n", trim($process->getOutput())))
+                ->filter(fn ($line) => str_contains($line, '::'))
+                ->count() >= self::MIN_TEST_COUNT;
+        }
```
sagalbot: Here's my suggested alternative for looking for the `/tests` directory. I think this will be less brittle with fewer edge cases. 

example where looking for `/tests` fails: 
- none of our apps have a `/tests` directory, they're all organized by domain, all tests are within subfolders
- pest allows you to colocate your tests, there might be no test files in `/tests`

This approach of calling `artisan test --list-tests` works with both pest and phpunit, and the `--list-tests` flag has been supported since PHPUnit 4.0. 

I also updated the test count to be a const, and changed it to two. Otherwise, installing boost into a fresh laravel app would prevent the test guidelines from being added.

ashleyhindle: I love this - we might have to use `PHP_BINARY` and set the CWD so `artisan` works, but this is a much better approach 🙌

billypoke: since you're already `collect()`ing the output of the process, you could wrap it with the `Str::` helper as follows:
```
$hasMinimumTests = Str::of($process->getOutput())
    ->trim()
    ->explode("\n")
    ->filter(fn ($line) => str_contains($line, '::'))
    ->count() >= self::MIN_TEST_COUNT;

sagalbot: @ashleyhindle updated to `PHP_BINARY` and set the `cwd` to `base_path()`

@billypoke thanks for that suggestion, much cleaner!

pushpak1300 (approved):
@ashleyhindle I have test this locally. 
 
LGTM 🚀

ashleyhindle:
Thanks @sagalbot, you're the best 🫶

PR #99, opened 2025-08-14
Fix: Prevent autoloading non class-like files during discovery to avoid "FatalError: Cannot redeclare function"

### LINKED ISSUE #67: Cannot install, duplicate function declaration
### Laravel Package Version

1.0.3

### Laravel Version

12.20.0

### PHP Version

8.3.15

### Database Driver & Version

_No response_

### Description

When running `php artisan boost:install` I get duplicate declaration of helper functions that are loading via composer

```
"autoload": {
        "psr-4": {
            "App\\": "app/",
            "Database\\Factories\\": "database/factories/",
            "Database\\Seeders\\": "database/seeders/"
        },
        "files": [
            "app/helpers.php"
        ]
    },
```

Cannot redeclare template_path() (previously declared in html/app/helpers.php:6)

### Steps To Reproduce

`composer require laravel/boost --dev`

Add helpers file to autoloader and place function inside.

run `php artisan boost:install`
Select PHPStorm, then Junie

zdearo:
Description Summary
- This PR resolves some points discussed in #67 
- Add a lightweight guard (fileHasClassLike) before calling class_exists during app discovery.
- Skip files that do not contain a class/interface/trait/enum to prevent unintended autoloading of helpers or function-only files.

Motivation
- class_exists triggered Composer’s autoload for helper files that only declare functions, which could lead to double-inclusion and a Fatal error: Cannot redeclare function.
- Discovery only needs class-like types (Models, Controllers, Enums), so loading helpers is unnecessary and risky.

Changes
- Introduce fileHasClassLike($path) with:
    - static in-memory cache per path;
    - cheap keyword pre-check to avoid tokenization when possible;
    - token-based detection (T_CLASS, T_INTERFACE, T_TRAIT, T_ENUM) to avoid false positives.

- Use this guard in the discovery loop to skip non class-like files before invoking class_exists.

Fixes #67

### DIFF
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -112,7 +112,7 @@ class GuidelineAssist
             return $cache[$path];
         }
 
-        $code = @file_get_contents($path);
+        $code = file_get_contents($path);
         if ($code === false) {
             return $cache[$path] = false;
         }
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -80,8 +80,14 @@ class GuidelineAssist
                 );
 
                 try {
+                    $path = $appPath.DIRECTORY_SEPARATOR.$relativePath;
+
+                    if (! $this->fileHasClassLike($path)) {
+                        continue;
+                    }
+
                     if (class_exists($className)) {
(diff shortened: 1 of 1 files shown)


ashleyhindle (approved):
Wonderful, thanks for finding & fixing this @zdearo, super helpful! 🫶

Issue #75, opened 2025-08-13
Can´t start it in PhpStorm


Hellcastle:
### Laravel Package Version

1.0.5

### Laravel Version

12.23.1

### PHP Version

8.3.24

### Description

When I went to enable all the new Laravel Boost features/connections for Junie in PHPStorm, the status for Herd and laravel-boost is "Cannot run program "php": CreateProcess error=2, The system cannot find the file specified."
I reinstalled Boost, but without success.

<img width="2230" height="1263" alt="Image" src="https://github.com/user-attachments/assets/5f21cc25-fd29-42a9-b6fb-62beed2fefe3" />

mcp.json

```java
{
    "mcpServers": {
        "laravel-boost": {
            "command": "php",
            "args": [
                "C:\\Users\\xxxx\\Herd\\project\\artisan",
                "boost:mcp"
            ]
        },
        "herd": {
            "command": "php",
            "args": [
                "C:/Users/xxxx/.config/herd/bin/herd-mcp.phar"
            ],
            "env": {
                "SITE_PATH": "C:\\Users\\xxxx\\Herd\\project"
            }
        }
    }
}
``` 

PHPStorm Version:
PhpStorm 2025.2
Windows 11
Herd 1.22.0

### Steps To Reproduce

Create blank Laravel 12 project with Herd, Livewire Stack, Laravels build-in authentication and Pest, no Volt
add boost via composer require laravel/boost --dev
php artisan boost:install with this settings:

<img width="2330" height="975" alt="Image" src="https://github.com/user-attachments/assets/07f88fa2-dc09-4f86-b3f0-4c0bb4cedf6a" />

ashleyhindle:
If you run `php -v` in the same terminal you installed Boost in, does it run ok?

I think PHPStorm has issues with PATHs, we might need to use an absolute path for PHP too 🤔

dinoluck:
> If you run `php -v` in the same terminal you installed Boost in, does it run ok?
> 
> I think PHPStorm has issues with PATHs, we might need to use an absolute path for PHP too 🤔

I found that if I escape it properly (C:\\path\\to\\herd\\php\\executable) it works, but then it hangs "starting" because the .phar itself is choking or something else. Even if you try to execute it from like PowerShell "php ...mcp.phar" it just hangs until you kill it.

Hellcastle:
Yes, here is the result:

[...]

Issue #74, opened 2025-08-13
Cannot run program "php": CreateProcess error=2, The system cannot find the file specified


dev-idkwhoami:
### Laravel Package Version

1.0.4

### Laravel Version

 12.23.1

### PHP Version

8.4.11

### Description

After freshly installing the laravel/boost package and running the boost:install command which finished successfull. I tried activating the MCP Server in PHPStorm's MCP Settings. But i got an error saying `Cannot run program "php": CreateProcess error=2, The system cannot find the file specified`.

I want to clarify they work as intended, within PHPStorm as well as in a normal console.

> [!IMPORTANT]
> I relatively quickly figured out that for `CreateProcess` doesn't like the "shortened" php so i changed the program to `php.bat` and then it could start the server successfully.

For completion, this is what my `.junie/mcp/mcp.json` file looks like
```json
{
    "mcpServers": {
        "laravel-boost": {
            "command": "php.bat",
            "args": [
                "B:\\zenni\\artisan",
                "boost:mcp"
            ]
        }
    }
}
```

I don't know whether that is something Jetbrains needs to fix or maybe the boost:install command can "work around" ?

I don't even know if everybody using herd with standalone php has that issue.

Anyway i decided to post this here so that others having the same issue might be able to solve it.

### Prerequisites:
Laravel Herd installed and PHP installed through Herd (PHP standalone). Might be important im on Windows 11.

### Steps:
1. Install laravel/boost composer package:
```shell
composer require laravel/boost --dev
```
2. Run boost:install command
```shell
php artisan boost:install
```
3. Navigate to PHPStorm Settings > Tools > Junie > MCP Settings and activate the laravel-boost server (click Apply)
4. The "Status" column will show red circle with exclamation mark. By clicking into the cell it will display the error.

dinoluck:
If you change "command" to "C:\\path\\to\\php\\executable" it will fix that error for you.

otavio-araujo:
> If you change "command" to "C:\path\to\php\executable" it will fix that error for you.

Worked for me.

ashleyhindle:
This should now be fixed for new installations using the latest Boost, marking as closed 🙏

Marcel is looking at Herd not starting in a different issue 👌

Issue #48, opened 2025-08-13
StandaloneCoroutine was cancelled


cdebattista:
### Laravel Package Version

10.48.25

### Laravel Version

10.48.25

### PHP Version

8.3.24

### Database Driver & Version

MariaDB

### Description

<img width="983" height="739" alt="Image" src="https://github.com/user-attachments/assets/9e116067-ac71-4970-ace9-56d8306500a8" />
<img width="983" height="739" alt="Image" src="https://github.com/user-attachments/assets/ec946153-db6e-4005-bc27-7a384e848d5b" />

### Steps To Reproduce

composer require laravel/boost --dev
php artisan boost:install

PhpStorm -> Junie

jobjen02:
See https://github.com/laravel/boost/issues/36#issuecomment-3184361793, might be related to Docker?

ghost:
I'm not in docker env

pushpak1300:
can you share complete exception trace ? @cdebattista

ghost:
StandaloneCoroutine was cancelled

[...]

PR #109, opened 2025-08-14
Fix PHPStorm using absolute paths

ashleyhindle:
PHPStorm is the only editor that requires absolute paths in its config. Right now we don't pass an absolute path to 'php' so people are having issues using Boost in PHPStorm.

- https://github.com/laravel/boost/issues/75
- https://github.com/laravel/boost/issues/74
- Maybe https://github.com/laravel/boost/issues/48

### Caveats
Absolute paths make it impossible to commit the resulting file for use within a team. We might need to update docs to say "Add to .gitignore and each dev manually adds until it's improved, sorry!" 😬


cord:
on macOs `whereis php` returns `/Users/cord/Library/Application Support/Herd/bin//php` as absolute path to php.

Looks like the blank in the path is causing issues?

ashleyhindle:
> on macOs `whereis php` returns `/Users/cord/Library/Application Support/Herd/bin//php` as absolute path to php.
> 
> Looks like the blank in the path is causing issues?

Ah yes, PHPStorm also can't handle spaces even if you escape them 🤔 Good point

calebdw:
@ashleyhindle, please review https://github.com/laravel/boost/pull/83 before this is merged as this will break how that detects and switches between `php` and `sail` as the primary command---just something to think about as you architect this

pushpak1300 (approved):

PR #113, opened 2025-08-14
Ensure guideline file content ends with a newline

balu-lt:
The initial guideline file is generated without a newline at the end.
When additional custom rules are added, code editors like PhpStorm automatically append a newline, causing unnecessary GIT diffs.

This PR ensures that the generated guideline content always ends with a newline, preventing unwanted diffs and following common text file conventions.


nikspyratos:
Would it be worth expanding this PR to include doing this change for the `mcp.json` files created/modified? I've come across this issue there too.

PR #118, opened 2025-08-14
Add Herd to default install options when detected

mpociot:
This PR adds the Herd MCP server to the list of default options once it gets detected.
This does not affect users who don't have Herd installed, but if Herd is installed, the user no longer has to manually enable it.

If you use Herd, you are most likely also going to use the Herd MCP server – at least I don't see a reason why someone would not want to enable it in this case.


RiadhKHEDHIRI:
🙏

pushpak1300 (approved):

PR #142, opened 2025-08-16
Add support for Vite CSP nonce

nckrtl:
Boost Browserlogger won't work when CSP is enabled with a Vite provided nonce. This commit adds the Vite nonce when its detected.

Boost is meant for local development and CSP is often applied in production only. Yet for testing purposes CSP can be applied in the local environment as well, making it useful for Boost to support CSP nonces as well. Also, when implementing CSP it will produce console errors, so Boost Browserlogger could make implementing CSP a lot easier as well.


ashleyhindle (approved):
💪 Thank you!

Issue #146, opened 2025-08-16
Support for ordering generated files


imliam:
Anecdotally, I have found that various LLMs (particularly Claude and Gemini) put a lot more weight on _early_ context they're given, eg. at the top of a guidelines file. Unfortunately, any context in `.ai/guidelines/*` gets appended to the _end_ of the generated guidelines file.

In my opinion, project-specific guidelines should come first so the developer has more control over what context the LLM is more likely to remember when performing any given task. I'd be okay. with them always being prepended to the guidelines list, but maybe a way to configure the order of the guidelines would be better for the ecosystem depending on use cases?

calebdw:
Along with this, I believe that there needs to be explicit verbiage that the LLM should always follow the user provided guidance over the boost instructions---I need to override some parts but I can't just modify the file directly because the file is overwritten on every install

PR #150, opened 2025-08-16
Put user-defined project guidelines at the top of the generated guidelines file

imliam:
Potentially solves #146 

Just raising this as an easy merge if you agree that user guidelines should take more precedence than the Boost presets, but happy to do another approach if liked


pushpak1300:
@imliam can you resolve the conflict ?

imliam:
Closing in favour of #332

PR #172, opened 2025-08-19
fix: defer InjectBoost middleware registration until app is booted

Sairahcaz:
This PR fixes an issue where the `InjectBoost` middleware was not reliably appended to the web middleware group in the old Kernel based app configuration.

By deferring the call with `$this->app->booted(...)`, we ensure the middleware is always appended to the final web group definition.


ashleyhindle:
Appreciate this @Sairahcaz, thanks!

PR #182, opened 2025-08-20
Clarify ListRoutes name parameter description for better tool calling

### LINKED ISSUE #181: Route list tool fails to match routes when using wildcards
### Laravel Package Version

1.0.16

### Laravel Version

12

### PHP Version

8.2

### System Info

Mac OS with Laravel Herd

### Description

When using the list-routes tool with wildcard criteria like name: "two-factor*" or path: "*/two-factor*" it fails to return any matching routes, even though relevant routes clearly exists.

### Steps To Reproduce

Use list-routes (MCP)(name: "two-factor*") or list-routes (MCP)(path: "*/two-factor*")

Observe the error:

ERROR  Your application doesn't have any routes matching the given criteria.

Expected behavior:

The tool should correctly match and return routes when using wildcards, similar to Laravel’s native artisan route:list.

Image

pushpak1300:
**Problem:** LLMs were attempting to use unsupported wildcard patterns like `name="*/two-factor/"` with the ListRoutes tool.

**Solution:** Updated parameter descriptions to explicitly clarify that wildcards are not supported and added examples for better clarity.

Fixes #181

### DIFF
--- a/tests/Feature/Mcp/Tools/ListRoutesTest.php
+++ b/tests/Feature/Mcp/Tools/ListRoutesTest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 use Illuminate\Support\Facades\Route;
 use Laravel\Boost\Mcp\Tools\ListRoutes;
-use Laravel\Mcp\Server\Tools\ToolResult;
 
 beforeEach(function () {
     Route::get('/admin/dashboard', function () {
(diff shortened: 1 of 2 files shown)


ashleyhindle (approved):
Should we also _remove_ asterisks from any input, so even when the AI agent gets it wrong, Boost still gets it right?

Issue #129, opened 2025-08-15
MCP continuously crashes with timeout errors when executing long running Tinker commands


mathiaswillburger:
### Laravel Package Version

1.0.17

### Laravel Version

12.24.0

### PHP Version

8.4

### System Info

Windows 11, VSCode

## Issue Description
The MCP server repeatedly crashes when executing Tinker commands, requiring manual restarts. The server consistently hits the 30-second PHP execution timeout limit and terminates with exit code 255.

The tinker command Sonnet is trying to execute is including various things like clearing the cache, creating a job, using a service to upload data, etc. So it takes some time to complete.

## Expected Behavior
The MCP server should handle Tinker commands without timing out and maintain a stable connection for extended development sessions.

## Suggested Solutions
I probably could update the PHP execution time locally, but I feel that this is something the user shouldnt worry about and could be done by boost.  --> **Increase PHP execution timeout**: Configure `max_execution_time` or use `set_time_limit(0)` for the MCP server process

## Actual Behavior
- Server starts successfully and discovers the tools
- After approximately 30 seconds of operation, the server crashes with a fatal PHP timeout error
- The process exits with code 255
- Manual restart is required to restore functionality
- This pattern repeats consistently

## Log Sample
```
2025-08-14 20:08:12.925 [info] Connection state: Running
2025-08-14 20:08:13.455 [info] Discovered 16 tools
2025-08-14 20:12:27.238 [warning] [server stderr] PHP Fatal error:  Maximum execution time of 30 seconds exceeded in C:\xxxxxxx\vendor\laravel\mcp\src\Server\Transport\StdioTransport.php on line 57
2025-08-14 20:12:27.246 [warning] Failed to parse message: "\r\n"
2025-08-14 20:12:27.246 [warning] Failed to parse message: "   Symfony\\Component\\ErrorHandler\\Error\\FatalError \r\n"
2025-08-14 20:12:27.247 [warning] Failed to parse message: "\r\n"
2025-08-14 20:12:27.247 [warning] Failed to parse message: "  Maximum execution time of 30 seconds exceeded\r\n"
2025-08-14 20:12:27.247 [warning] Failed to parse message: "\r\n"
2025-08-14 20:12:27.247 [warning] Failed to parse message: "  at C:\\xxxxxxx\\vendor\\laravel\\mcp\\src\\Server\\Transport\\StdioTransport.php:57\r\n"
[...]

PR #184, opened 2025-08-21
Fix MCP crashes on Windows with process isolation

andreilungeanu:
A new PR on same subject after some brainstorming, is not about the timeout value...
Fix for #129 

## Problem
The Tinker tool was crashing the entire MCP server on Windows due to PHP's `set_time_limit()` causing uncatchable fatal errors when timeouts occur. This made the MCP server unstable and unusable on Windows systems.

## Root Cause
- **Windows limitation**: `set_time_limit()` + infinite loops = fatal error crash (cannot be caught)
- **PCNTL unavailable**: Windows doesn't support PCNTL signals for timeout handling
- **No graceful recovery**: Server process dies completely, affecting all connected clients

## Solution
Implemented **process isolation by default** with platform-aware fallbacks:

### Changes

1. **Process Isolation Config** (`config/boost.php`)
   ```php
   'process_isolation' => [
       'enabled' => env('BOOST_PROCESS_ISOLATION', true),  // ON by default
       'timeout' => env('BOOST_PROCESS_TIMEOUT', 180),
   ]
   ```

2. **Enhanced Tinker Tool** (`src/Mcp/Tools/Tinker.php`)
   - **Primary**: Relies on ToolExecutor process isolation (Windows-safe)
   - **Fallback**: Uses `set_time_limit()` only when isolation disabled (Unix systems)
   - **PCNTL support**: Additional protection on Unix systems
   - **Proper cleanup**: Finally blocks ensure resource cleanup

3. **Platform-Aware Tests** (`tests/Feature/Mcp/Tools/TinkerTest.php`)
   - Automatically skips timeout tests on Windows (PCNTL unavailable)
   - Tests both isolation enabled/disabled scenarios

## Configuration
```bash
# Default (recommended for Windows)
BOOST_PROCESS_ISOLATION=true

# Performance mode (Unix only - risky on Windows)
BOOST_PROCESS_ISOLATION=false
```

## 📊 Platform Behavior

| Platform | Process Isolation | Timeout Method | Risk Level |
|----------|------------------|----------------|------------|
| **Windows** | ✅ Enabled (required) | ToolExecutor isolation | ✅ Safe |
| **Windows** | ❌ Disabled | `set_time_limit()` | ⚠️ **CRASHES** |
| **Unix/Linux** | ✅ Enabled | ToolExecutor isolation | ✅ Safe |
| **Unix/Linux** | ❌ Disabled | `set_time_limit()` + PCNTL | ✅ Safe |

**Bottom line**: Windows users can now use Tinker without fear of crashing the MCP server.

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -29,4 +29,24 @@ return [
 
     'browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Process Isolation for Tools
+    |--------------------------------------------------------------------------
+    |
+    | When enabled, tools like Tinker will run in isolated processes to prevent
+    | timeouts from affecting the main MCP server. This provides better stability
+    | and protection against runaway code execution.
+    |
+    | Platform Considerations:
+    | - Windows: Process isolation is strongly recommended as timeout errors
+    |   cause uncatchable fatal crashes when disabled
+    | - Unix/Linux: Can be safely disabled for better performance as PCNTL
+    |   provides graceful timeout handling with catchable exceptions
+    */
+
+    'process_isolation' => [
+        'enabled' => env('BOOST_PROCESS_ISOLATION', true),  // Default ON for cross-platform safety
+        'timeout' => env('BOOST_PROCESS_TIMEOUT', 180),     // 3 minutes default
+    ]
 ];
--- a/tests/Feature/Mcp/Tools/TinkerTest.php
+++ b/tests/Feature/Mcp/Tools/TinkerTest.php
@@ -174,6 +174,14 @@ test('uses default timeout when not specified', function () {
 });
 
 test('times out when code takes too long', function () {
+    // Skip if PCNTL functions are not available
+    if (!function_exists('pcntl_async_signals') || !function_exists('pcntl_signal')) {
+        $this->markTestSkipped('PCNTL functions not available for timeout testing');
+    }
+
+    // Disable process isolation for this test
+    config(['boost.process_isolation.enabled' => false]);
+
     $tool = new Tinker;
 
     // Code that will take more than 1 second to execute
@@ -191,5 +199,23 @@ test('times out when code takes too long', function () {
 
     $data = getToolResultData($result);
     expect($data)->toHaveKey('error')
-        ->and($data['error'])->toMatch('/(Maximum execution time|Code execution timed out)/');
+        ->and($data['error'])->toMatch('/(Maximum execution time|Code execution timed out|Fatal error)/');
+});
+
+test('relies on process isolation for timeout protection on Windows', function () {
+    // This test documents that Windows relies on process isolation for safe timeout handling
+    if (PHP_OS_FAMILY !== 'Windows') {
+        $this->markTestSkipped('Windows-specific timeout behavior test');
+    }
+
+    // Process isolation enabled (recommended for Windows)
+    config(['boost.process_isolation.enabled' => true]);
+
+    $tool = new Tinker;
+    $result = $tool->handle(['code' => 'return "Windows: process isolation provides safe timeout protection";']);
+
+    expect($result)->toBeInstanceOf(ToolResult::class);
+
+    $data = getToolResultData($result);
+    expect($data['result'])->toBe('Windows: process isolation provides safe timeout protection');
 });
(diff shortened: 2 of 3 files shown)


ahosker:
Thanks for this, it realy does keep crashing for me on windows with timeout messages.

andreilungeanu:
> Thanks for this, it realy does keep crashing for me on windows with timeout messages.

After applying this PR, turn on process isolation in boost.php, ever since I did that, no more crashes from the timeout issue on Windows in VSCode. (if you had the boost.php config published before this change, remove it or add manually new options) 

More recently, I also set isolation to be enabled by default, even if your old config didn’t include that option. [75d215f](https://github.com/laravel/boost/pull/184/commits/75d215f264344f9d0a27e62cdde18c9d913aa8a1)

config/boost.php
```diff
@@ -29,4 +29,24 @@
 
     'browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
 
+    /*
```
ashleyhindle: We don't want to add the config options here at the minute - we'll document them in the README/similar soon instead 👌

tests/Feature/Mcp/Tools/TinkerTest.php
```diff
@@ -174,6 +174,14 @@ function getToolResultData(ToolResult $result): array
 });
 
 test('times out when code takes too long', function () {
+    // Skip if PCNTL functions are not available
```
ashleyhindle: Can you bring this file up to date to fit this in?

We recently made some changes so this is in conflict atm 👌

andreilungeanu: Sure, Ashley. Thank for reviewing this, I will update this asap

andreilungeanu:
Since parts of this PR were already implemented in [feat: default process isolation to true](https://github.com/laravel/boost/commit/75a4cc8765b924bb642cddc285418c17d132ed45), things are much simpler now. With the config section removed as requested and the fallback changed from `false` to `true`, the code relies on `config('boost.process_isolation.enabled', true)` which always returns `true`. 

Windows users get safe process isolation by default, and this PR now simply focuses on platform-aware timeout handling and test coverage.

ashleyhindle:
> Since parts of this PR were already implemented in [feat: default process isolation to true](https://github.com/laravel/boost/commit/75a4cc8765b924bb642cddc285418c17d132ed45), things are much simpler now. With the config section removed as requested and the fallback changed from `false` to `true`, the code relies on `config('boost.process_isolation.enabled', true)` which always returns `true`.
> 
> Windows users get safe process isolation by default, and this PR now simply focuses on platform-aware timeout handling and test coverage.

🤔  What do you think about us removing the process isolation configuration _entirely_, so we can _always_ rely on the subprocess timeout management?

Then in `ToolExecutor` use `$arguments['timeout'] ?? 180` as the timeout?

**Thinking**
Originally the config was there in case the process isolation had issues, but it's been working really well for me and I don't see why it should ever be disabled.

Right now, as you documented, if somebody disables process isolation, then we use `set_time_limit` in Tinker and their code takes too long, the entire MCP process dies. 

**Two options**
Fundamentally the Tinker tool should always run the eval'ed code in a fork or subprocess to protected against that, so we could add that into the `Tinker` tool _additionally_, or just _only_ work with process isolation because it's great and works?

I recognise this is a departure from your original plan, but it solves the same issue and simplifies the Tinker tool a lot.

andreilungeanu:
@ashleyhindle  you're right  - that was actually my first instinct too, but I second-guessed myself thinking the changes might be too brutal for the initial PR. The current approach of always using subprocess with getTimeout() method that extracts $arguments['timeout'] ?? 180 and clamps it between 1-600 seconds eliminates the crash risk completely while dramatically simplifying the architecture. 

One thing I'd also suggest is reconsidering the hardcoded ini_set('memory_limit', '128M') in Tinker - for large projects this might be too restrictive. It should probably be configurable or at least doubled to 256M to ensure it works properly across different project sizes without running into memory issues during code execution.

src/Mcp/ToolExecutor.php
```diff
@@ -71,30 +67,11 @@ protected function executeInProcess(string $toolClass, array $arguments): ToolRe
...
-        }
-    }
-
-    protected function shouldUseProcessIsolation(): bool
+    protected function getTimeout(array $arguments): int
     {
-        if (app()->environment('testing')) {
-            return false;
-        }
+        $timeout = (int) ($arguments['timeout'] ?? 180);
```
ashleyhindle, line 72: I wonder if it would make sense to use `Number::clamp` here to ensure it's between 1 and 600 seconds, say?

Some models might set a timeout of `-1` or 86400, and I'm not sure those tinker commands are beneficial to the user if they run that long and we should allow the AI agent to fail-early so it can think of a more appropriate choice

ashleyhindle: Ah, that's the line below 😁

ashleyhindle (approved):
Great change, thank you @andreilungeanu, and appreciate you going the extra mile with the latest refactor 🙌

Issue #176, opened 2025-08-20
Tinker tool crashes mcp due to memory usage


mortenscheel:
### Laravel Package Version

v1.0.18

### Laravel Version

v12.19.3

### PHP Version

8.3.24

### System Info

macOS Sequoia 5.6 (24G84), php from homebrew. mcp used from claude code

### Description

⏺ laravel-boost - tinker (MCP)(code: "echo 123;")
  ⎿  Error: Inline tool execution failed: Failed to set memory limit to 134217728 bytes (Current memory usage is 200278016 bytes)

This happens in Tinker.php when it tries to set memory_limit to `128M`.

I've tried to debug this by logging the memory usage when Tinker.php's handle method is called.

Calling the tool manually, the memory usage is `56 MB`, number of declared classes: `1304`
```shell
php artisan boost:execute-tool 'Laravel\Boost\Mcp\Tools\Tinker' eyJjb2RlIjoiZWNobyAxMjM7In0=
```
When the same command sent from Claude Code, the memory usage is `191 MB`, number of declared classes: `4701`

I've also tried debugging with `mcp:inspector`, but the results are the same.

Note: this is happening in a very large Laravel application with ~400 models and a total of 4626 php files in the app/ folder.

### Steps To Reproduce

Run the tinker mcp tool in a very large Laravel project.

nikspyratos:
Does this keep happening if you actually increase the memory limit?

Purely off comparing the `execute-tool` vs actual `boost:mcp` commands, there's a lot more going on under the hood when running the MCP server. 

Compare the [tool execution command](https://github.com/laravel/boost/blob/main/src/Console/ExecuteToolCommand.php) vs the [underlying MCP server process handling](https://github.com/laravel/mcp/blob/main/src/Server.php). Plain executing the tool command once off is going to be pulling in a lot less app context as it's not doing any boot up & continuous running processes that may involve pulling in the rest of your code base into the process.

My 2c: Ashley would know better on the inner workings and how resource usage would be impacted this way, but in general it's not surprising that such a humongous project (yes, 4000+ files in `app` and 400 models **_is_** a very large project) would have higher baseline memory requirements just to load everything in.

kavanpancholi:
I have the same issue.

And the issue is that in the Tinker Mcp tool, it has hardcode limit set [here](https://github.com/laravel/boost/blob/main/src/Mcp/Tools/Tinker.php#L47)

It should be configurable from a config file

pushpak1300:
[...]

PR #196, opened 2025-08-25
Add config to modify tinker memory limit

kavanpancholi:
Fixes: https://github.com/laravel/boost/issues/176

On a large project, it throws memory limit error while running tinker mcp tool.
image

And since currently there is no option to configure memory limit, it is hard to fix it.
So, this PR adds config variable to update the memory limit.


taylorotwell:
The doc block will need proper formatting before it can be merged.

PR #203, opened 2025-08-27
Fix: MCP Environment Variable Caching Issue

### LINKED ISSUE #130: Boost doesn't detect environment changes
### Laravel Package Version

1.0.17

### Laravel Version

12.24.0

### PHP Version

8.4.11

### System Info

macOs 15.6 with Herd 1.22.0, Cursor 1.4.5

### Description

I installed Boost with a vanilla Laravel install using the sqlite driver. I changed to the mysql driver but the MCP server in Cursor continued returning sqlite as the default database driver both when ran using the get-config tool and when running `return config("database.default");` via Tinker.

The solution was to restart Cursor. Seems that either Boost or Cursor are caching the results from the MCP calls. I'm reporting it here in case this is a Boost configuration.

### Steps To Reproduce

1. Install a vanilla Laravel app
2. Install Boost
3. Change laravel to use a mysql connection
4. Ask the Cursor Agent to find out which is the default database connection (or ask it to create a new table).

It will attempt to use the sqlite connection in its generated commands or will try to verify that it's work is complete by querying the wrong database driver.

andreilungeanu:
# Problem
MCP server runs as a long-lived process that inherits environment variables at startup. When users change .env values, child processes spawned for tool execution inherit the parent's stale environment variables instead of reading fresh values from the updated .env file.

# Root Cause
MCP server loads .env values into $_ENV at startup
Child processes inherit parent's environment variables
Laravel's LoadEnvironmentVariables doesn't re-read .env if variables already exist in environment
Process isolation alone doesn't solve this due to environment inheritance

# Solution
Refresh environment variables ExecuteToolCommand 
Fix #130

### DIFF
--- a/src/Console/ExecuteToolCommand.php
+++ b/src/Console/ExecuteToolCommand.php
@@ -18,6 +18,9 @@ class ExecuteToolCommand extends Command
 
     public function handle(): int
     {
+        // Refresh environment variables
+        $this->refreshEnvironmentVariables();
+
         $toolClass = $this->argument('tool');
         $argumentsEncoded = $this->argument('arguments');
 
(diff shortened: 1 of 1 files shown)


ashleyhindle:
Apologies @andreilungeanu , didn't see this and just merged #217 with a similar fix 👌

Thanks for PRing, will keep a closer eye on your future PRs!

andreilungeanu:
Thanks @ashleyhindle  No worries at all — glad the issue is fixed either way! Happy to contribute again if I spot anything else.

Issue #107, opened 2025-08-14
Existing server configuration overwritten in mcp.json


amadeann:
### Laravel Package Version

1.0.11

### Laravel Version

11.44.0

### PHP Version

8.2.9

### System Info

Windows 11 + WSL2 + Laravel Sail + VSCode (with devcontainer)

### Description

Existing "servers" configuration such as:

```
{
	"servers": {
		"mysql": {
			"type": "stdio",
			"command": "npx",
			"args": [
				"@benborla29/mcp-server-mysql@2.0.3"
			],
			"env": {
				"MYSQL_HOST": "${input:mysql_host}",
				"MYSQL_PORT": "${input:mysql_port}",
				"MYSQL_USER": "${input:mysql_user}",
				"MYSQL_PASS": "${input:mysql_pass}",
				"MYSQL_DB": "${input:mysql_db}"
			}
		}
	},
	"inputs": [
		{
			"id": "mysql_host",
			"type": "promptString",
			"description": "MySQL Hostname"
		},
		{
			"id": "mysql_port",
			"type": "promptString",
			"description": "MySQL Port"
		},
		{
			"id": "mysql_user",
			"type": "promptString",a
			"description": "MySQL Username"
		},
		{
			"id": "mysql_pass",
			"type": "promptString",
			"description": "MySQL Password"
		},
		{
			"id": "mysql_db",
			"type": "promptString",
			"description": "MySQL Database Name (leave blank for multi-DB mode)"
		}
	]
}
```

gets overwritten on executing `php artisan sail:install`.

The culprit of this bug is likely here:
https://github.com/laravel/boost/blob/44f6eff403dd9a451ef900e9ddecb19e99463c97/src/Install/CodeEnvironment/CodeEnvironment.php#L180

### Steps To Reproduce

1. Create a new Laravel project in Laravel Sail on Windows 11 with WSL2
2. Open the project in VS Code with GitHub copilot, in a devcontainer
3. Add any MCP server, such as "@benborla29/mcp-server-mysql@2.0.3", and make sure the mcp.json file has an entry under "servers" key
4. Run "php artisan boost:install", and select "VS Code" in the first step as your editor
5. Previous value of "servers" in mcp.json gets replaced

pushpak1300:
@amadeann,

I tried reproducing the issue by creating a fresh Laravel 11 installation and adding a `.vscode/mcp.json` file with the following content:
```json
{
    "servers": {
        "mysql": {
            "type": "stdio",
[...]

PR #204, opened 2025-08-27
feat: add robust MCP file configuration writer

ashleyhindle:
Adds FileWriter class that 'intelligently' handles plain JSON files, and JSON5 style files which VS Code supports.

- Detects & preserves JSON 5 features (comments, trailing commas)
- 'Smart' injects based on existing keys and indents

Fixes https://github.com/laravel/boost/issues/107

--

We keep the basic `json_decode -> data_set -> json_encode -> write` process for 'plain' JSON files, and only fallback to complex injection when we detect a JSON5 style file. This allows people to keep their comments in `mcp.json` and we won't overwrite them, they'll stay retained, and we'll inject our config at the exact right place.


src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+        $baseIndent = str_repeat(' ', $baseIndent);
+
+        foreach ($lines as $i => $line) {
+            if ($i === 0) {
+                $indentedLines[] = $baseIndent.'"'.$key.'": '.$line;
+            } else {
+                $indentedLines[] = $baseIndent.$line;
+            }
+        }
+
```
pushpak1300, line 204: can it be simplify by this.

```php
collect($lines)->map(fn($line, $index) => $index 
    ? $indentationPrefix . '"' . $serverName . '": ' . $line
    : $indentationPrefix . $line)
```

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+        return "\n".implode("\n", $indentedLines);
+    }
+
+    protected function needsCommaAfterBrace(string $content, int $bracePosition): bool
+    {
+        $afterBrace = substr($content, $bracePosition + 1);
+        $trimmed = preg_replace('/^\s*(?:\/\/.*$)?/m', '', $afterBrace);
+
+        // If next char is } (empty object) or nothing, no comma needed
+        return ! empty($trimmed) && $trimmed[0] !== '}';
```
pushpak1300: Can be simplified with https://laravel.com/docs/12.x/strings#method-str-doesnt-start-with
```php
return filled($trimmed) && Str::doesntStartWith($trimmed, '}');
```

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+        // Get content between opening and closing braces
+        $innerContent = substr($content, $openBracePos + 1, $closeBracePos - $openBracePos - 1);
+
+        // Skip whitespace and comments to find last meaningful character
+        $trimmed = preg_replace('/\s+|\/\/.*$/m', '', $innerContent);
+
+        // If empty or ends with opening brace, no comma needed
+        if (empty($trimmed) || substr($trimmed, -1) === '{') {
+            return false;
+        }
```
pushpak1300: Use blank() and Str::endsWith():
```php
if (blank($trimmed) || Str::endsWith($trimmed, '{')) {
```

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    protected function readFile(): string
+    {
+        return File::get($this->filePath);
+    }
+
+    protected function writeFile(string $content): bool
+    {
+        return File::put($this->filePath, $content) !== false;
```
pushpak1300, line 408: we can use file operations with rescue() helper for better error handling:

```php
return rescue(
    fn() => File::put($this->filePath, $content) !== false,
    false
);
```

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+
+    public function save(): bool
+    {
+        $this->ensureDirectoryExists();
+
+        if ($this->shouldWriteNew()) {
+            return $this->createNewFile();
+        }
+
+        $content = $this->readFile();
```
pushpak1300, line 57: This methods like shouldWriteNew(), createNewFile(), readFile(), ensureDirectoryExists() are single-use. we can definitely inline them for cleaner approach. 

eg. ```$shouldCreateNewFile = !File::exists($this->filePath) || File::size($this->filePath) < 3;```

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+            return $this->updatePlainJsonFile($content);
+        }
+
+        return $this->updateJson5File($content);
+    }
+
+    protected function updatePlainJsonFile(string $content): bool
+    {
+        $config = json_decode($content, true);
+        if ($config === null) {
```
pushpak1300: blank() helper can be used instead of null check. what do you think ?

src/Install/Mcp/FileWriter.php
```diff
@@ -0,0 +1,384 @@
...
+    {
+        return ! $this->fileExists() || File::size($this->filePath) < 3; // To account for files that are just `{}`
+    }
+
+    protected function readFile(): string
+    {
+        return File::get($this->filePath);
+    }
+
+    protected function writeFile(string $content): bool
```
pushpak1300, line 406: we can definitely consider them making inline with proper variable name as they are only used once. what say ?

tests/fixtures/mcp-no-configkey.json5
```diff
@@ -0,0 +1,14 @@
...
+            "command": "existing-cmd"
+        }
+    },
+    "inputs": [
+        {
+            "id": "test_input",
+            "type": "promptString"
+        }
+    ]
+}
```
pushpak1300: nit: is the file without newline is intentional ?

src/Install/Mcp/FileWriter.php
ashleyhindle: Ah yea when I was updating these lines I was upset with it - let me take a look

ashleyhindle: Prefer this way https://github.com/laravel/boost/pull/204/commits/283ee399717133ef646ee0b1094853f1d5143f05

src/Install/Mcp/FileWriter.php
ashleyhindle: Which exceptions will be thrown by `File::put` for `rescue()` to be valuable here?

src/Install/Mcp/FileWriter.php
ashleyhindle: I'm happy with small methods that designate what they do, I don't think methods only exist for re-use, they're useful for containment & the naming helping it be cleaner.

`shouldWriteNew()` is _much_ cleaner than a long ternary in my mind

src/Install/Mcp/FileWriter.php
ashleyhindle: This is more a check for if there's an error so we shouldn't continue, so it should probably check last error or use JSON_THROW_ON_ERROR

src/Install/Mcp/FileWriter.php
pushpak1300: My bad. I think this is fine.

src/Install/Mcp/FileWriter.php
ashleyhindle: Changed to use json_last_error for detecting errors: https://github.com/laravel/boost/pull/204/commits/96933546a90fb2c69a9e53498a66703ba39159ce

src/Install/Mcp/FileWriter.php
ashleyhindle: Not a fan of this atm 👌 Couldn't live with that (that was how I originally had them and it made it messier)

tests/fixtures/mcp-no-configkey.json5
ashleyhindle: Nope! 😁 Changed in last commit

src/Install/Mcp/FileWriter.php
ashleyhindle: Nice, I've got to up my Str:: game!

Ah actually, is this available in Laravel 10? Same with `filled/blank/`? I'll go with `!Str::startsWith` instead

pushpak1300 (approved):

PR #207, opened 2025-08-27
Add eslint and prettier rules

laserhybiz:
These packages are used in the starter kits, and adding them will help ensure the code is linted and formatted correctly

### DIFF
--- a/.ai/eslint/core.blade.php
+++ b/.ai/eslint/core.blade.php
@@ -1,0 +1,4 @@
+## ESLint Code Linter
+
+- You must run `npx eslint --fix .` before finalizing changes to ensure your code adheres to the project's coding standards.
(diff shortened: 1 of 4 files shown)


pushpak1300:
We’re still working out how to add hooks so external packages can supply their own guidelines, but for now it can be extended with custom guidelines. I’ll let @ashleyhindle take a look and see if this is something needs to be part of the boost.

taylorotwell:
Thanks for your pull request to Laravel!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include. 

If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

PR #219, opened 2025-08-28
Allow guideline overriding

ashleyhindle:
- Linear: https://linear.app/laravel/issue/AI-96/allow-guideline-exclusion-and-overriding

# What & Why
Some people only have Inertia in part of their project, they don’t want Boost's guidelines recommending to create every new page with Inertia.

They should be able to put a file in `.ai/guidelines/inertia-laravel/core.blade.php` that we use instead of Boost's.

This opens up the door for a lot of customisation.

- Update README.md
- If overridden file exists, use that instead of Boost's
    - even/especially if it's empty.

# How I tested
1. Ran the tests
2. Ran `boost:install` in a demo project with custom guidelines and verified the guideline files include the overridden files, not the Boost files.

This makes sure custom guidelines still work, and overriding Boost guidelines works.
```shell
❯ tree .ai
.ai
└── guidelines
    ├── enforce-tests.blade.php
    ├── livewire
    │   └── 3
    │       └── core.blade.php
    └── tdd.blade.php
```

# Notes/thoughts
  - There's a future task to enable publishing guidelines to modify. And _possibly_ a future task to allow excluding guidelines in the config.
  - I think we should soon move Boost guidelines into `.ai/guidelines/` in the package too, to match the custom guidelines, so they're the same.


pushpak1300 (approved):
LGTM !  🚀 
Tested locally with custom guidelines. works as expected.
Thanks.

PR #224, opened 2025-08-29
Fix Laravel 10 bootstrap/app.php Documentation Inaccuracy

paulschoeman:
### Problem

The Laravel 10 core guidelines in `.ai/laravel/10/core.blade.php` contained an inaccurate statement claiming "There is no `bootstrap/app.php` application configuration in Laravel 10". This is incorrect and misleading to Boost users working with Laravel 10 applications.

### What This Fixes

**Before (Incorrect):**

```
- There is no `bootstrap/app.php` application configuration in Laravel 10:
```

**After (Accurate):**

```
- Laravel 10 has a `bootstrap/app.php` file that creates the application instance and binds kernel contracts, but does not use it for application configuration like Laravel 11:
```

### Impact on End Users

-   **Prevents Confusion**: Boost users on Laravel 10 will no longer be misled into thinking the `bootstrap/app.php` file doesn't exist
-   **Accurate Guidance**: Users will understand that while the file exists, it serves a different purpose than Laravel 11's version
-   **Better Decision Making**: AI assistants using these guidelines will provide more accurate advice about Laravel 10 application structure

### Technical Details

Laravel 10 **does** have a `bootstrap/app.php` file, but it serves a foundational role:

-   Creates the application instance
-   Binds kernel contracts
-   **Does not** handle application configuration (unlike Laravel 11)

The configuration responsibilities remain in their traditional Laravel 10 locations:

-   Middleware registration: `app/Http/Kernel.php`
-   Exception handling: `app/Exceptions/Handler.php`
-   Console commands: `app/Console/Kernel.php`

### Why This Matters

This inaccuracy could lead to:

-   Developers incorrectly assuming they need to create a `bootstrap/app.php` file
-   Confusion when examining existing Laravel 10 codebases
-   AI assistants providing incorrect structural advice

### Backward Compatibility

✅ This change does not break any existing functionality - it only corrects documentation to match Laravel 10's actual file structure.

### Testing

-   [x] Verified against actual Laravel 10 application structure
-   [x] Confirmed the file exists and serves the described purpose
-   [x] Ensures guidelines accurately reflect framework behavior

This fix ensures Boost provides accurate, version-specific guidance that aligns with Laravel 10's actual architecture.

### DIFF
--- a/.ai/fluxui-pro/core.blade.php
+++ b/.ai/fluxui-pro/core.blade.php
@@ -16,5 +16,5 @@
 This is correct as of Boost installation, but there may be additional components within the codebase.
 
 <available-flux-components>
-accordion, autocomplete, avatar, badge, brand, breadcrumbs, button, calendar, callout, card, chart, checkbox, command, context, date-picker, dropdown, editor, field, heading, icon, input, modal, navbar, pagination, popover, profile, radio, select, separator, switch, table, tabs, text, textarea, toast, tooltip
+accordion, autocomplete, avatar, badge, brand, breadcrumbs, button, calendar, callout, card, chart, checkbox, command, context, date-picker, dropdown, editor, field, heading, file upload, icon, input, modal, navbar, pagination, pillbox, popover, profile, radio, select, separator, switch, table, tabs, text, textarea, toast, tooltip
 </available-flux-components>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,42 @@
 # Release Notes
 
-## [Unreleased](https://github.com/laravel/boost/compare/v1.2.1...main)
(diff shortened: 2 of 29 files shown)


ashleyhindle (approved):
Thank you 🙌

paulschoeman:
Hi @ashleyhindle 👋

I see we have a static analysis failure after main branch was merged in today.

It looks like this is due to the `laravel/mcp` package update last week where `ServerContext::tools()` is now parameter-less. 

Should be a quick fix to remove that argument from Line 41 of `src/Mcp/Methods/CallToolWithExecutor.php`:

```
->tools($request->toRequest())
```

I can do that in a separate PR if that's cleaner and helps?

PR #225, opened 2025-08-29
Streamline ToolResult assertions in tests 

pushpak1300:
This MR refactored all test files in `/tests/Feature/Mcp/Tools/` to use custom Pest expectations instead of repetitive assertion patterns.

### DIFF
--- a/tests/Feature/Mcp/Tools/BrowserLogsTest.php
+++ b/tests/Feature/Mcp/Tools/BrowserLogsTest.php
@@ -205,8 +205,8 @@ HTML;
     $content = $result->getContent();
     expect($content)->toContain('browser-logger-active')
         ->and($content)->toContain('</head>')
+        // Should not inject twice
         ->and(substr_count($content, 'browser-logger-active'))->toBe(1);
-    // Should not inject twice
 });
 
 test('InjectBoost middleware does not inject into non-HTML responses', function () {
--- a/tests/Feature/Mcp/Tools/BrowserLogsTest.php
+++ b/tests/Feature/Mcp/Tools/BrowserLogsTest.php
@@ -10,7 +10,6 @@ use Illuminate\Support\Facades\Route;
 use Laravel\Boost\Mcp\Tools\BrowserLogs;
 use Laravel\Boost\Middleware\InjectBoost;
 use Laravel\Boost\Services\BrowserLogger;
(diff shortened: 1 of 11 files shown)


tests/Feature/Mcp/Tools/BrowserLogsTest.php
```diff
@@ -36,47 +35,38 @@
...
 
-    expect($result)->toBeInstanceOf(ToolResult::class);
+    expect($result)->isToolResult()
+        ->toolHasNoError()
+        ->toolTextContains('browser.WARNING: Warning message', 'browser.ERROR: JavaScript error occurred')
+        ->toolTextDoesNotContain('browser.DEBUG: console log message');
 
     $data = $result->toArray();
-    expect($data['isError'])->toBeFalse();
     expect($data['content'][0]['type'])->toBe('text');
```
ashleyhindle, line 44: Is this a common assertion that should be made easy in `Pest.php`? It stands out that we do `result->toArray()` then check this

tests/Feature/Mcp/Tools/DatabaseConnectionsTest.php
```diff
@@ -18,16 +17,15 @@
...
-
-    $content = json_decode($data['content'][0]['text'], true);
-    expect($content['default_connection'])->toBe('mysql');
-    expect($content['connections'])->toHaveCount(3);
-    expect($content['connections'])->toContain('mysql');
-    expect($content['connections'])->toContain('pgsql');
-    expect($content['connections'])->toContain('sqlite');
+    expect($result)->isToolResult()
+        ->toolHasNoError()
+        ->toolJsonContent(function ($content) {
```
ashleyhindle: Should we just be using matches array? https://pestphp.com/docs/expectations#content-tomatcharrayarray

pushpak1300: I added one extra custom expectation to handle this

tests/Feature/Mcp/Tools/BrowserLogsTest.php
pushpak1300: This is not a widely used assertion I don't think we need to extract this in separate one.

ashleyhindle:
Very nice 💪

PR #229, opened 2025-08-30
Add strict types declaration in Inertia.php

felipeArnold:
The declare(strict_types=1); has been added to the inertia.php file to satisfy the architecture test regarding the use of strict types.


ashleyhindle (approved):
Thank you @felipeArnold 🙌

PR #261, opened 2025-09-16
fix: remove stray parenthesis in lifecycle hook guidance in livewire guidelines

mohammedyh:
Correct grammatical error by removing a stray closing parenthesis from the Livewire guidance on using lifecycle hooks


pushpak1300:
Thanks !

PR #263, opened 2025-09-17
[WIP] Makes boost use MCP 0.2

nunomaduro:


src/Console/ExecuteToolCommand.php
```diff
@@ -36,22 +39,34 @@ public function handle(): int
...
+        $request = new Request($arguments ?? []);
 
-            return 0;
+        try {
+            /** @var Response $response */
+            $response = $tool->handle($request);
+        } catch (Throwable $e) {
+            $errorResult = Response::error("Tool execution failed (E_THROWABLE): {$e->getMessage()}");
+
+            $this->error(json_encode([
```
ashleyhindle, line 53: I _think_ this might break @nunomaduro 

ExecuteTool command is ran on the terminal via a subprocess then its output json decoded, if 'error' outputs anything around the json_encode that'll break

src/Mcp/Tools/DatabaseConnections.php
```diff
@@ -4,32 +4,38 @@
...
+     */
+    protected string $description = 'List the configured database connection names for this application.';
 
-    public function schema(ToolInputSchema $schema): ToolInputSchema
+    /**
+     * Get the tool's input schema.
+     *
+     * @return array<string, JsonSchema>
+     */
+    public function schema(JsonSchema $schema): array
```
ashleyhindle, line 26: Can this be completely removed? I imagine the default is empty array right?

src/Mcp/Tools/DatabaseSchema.php
```diff
@@ -5,52 +5,55 @@
...
     /**
-     * @param array<string> $arguments
+     * Handle the tool request.
      */
-    public function handle(array $arguments): ToolResult
+    public function handle(Request $request): Response
     {
-        $connection = $arguments['database'] ?? config('database.default');
-        $filter = $arguments['filter'] ?? '';
+        $connection = $request->get('database') ?? config('database.default');
```
ashleyhindle, line 48: Can this use `$request->get('database', config('database.default'));`?

Does `->get()` support 'default'?

src/Mcp/Tools/ReportFeedback.php
```diff
@@ -19,34 +20,36 @@ public function description(): string
...
+     *
+     * @return array<string, JsonSchema>
+     */
+    public function schema(JsonSchema $schema): array
     {
-        return $schema
-            ->string('feedback')
-            ->description('Detailed feedback from the user on what would make Boost, or their experience with Laravel, better. Ask the user for more details if ambiguous or unclear.')
-            ->required();
+        return [];
```
ashleyhindle, line 30: We lost this input @nunomaduro - we'll need this 👌

src/Mcp/Tools/SearchDocs.php
```diff
@@ -20,46 +21,41 @@ public function __construct(protected Roster $roster)
...
-            ->integer('token_limit')
-            ->description('Maximum number of tokens to return in the response. Defaults to 10,000 tokens, maximum 1,000,000 tokens.')
-            ->optional();
+        return [
+            'queries' => $schema->array()
+                ->items($schema->string())
+                ->description('List of queries to perform, pass multiple if you aren\'t sure if it is "toggle" or "switch", for example')
+                ->required(),
+            'packages' => $schema->array()
+                ->items($schema->string())
```
ashleyhindle, line 42: Can the items have a description too @nunomaduro ? This is missing now

src/Mcp/Tools/Tinker.php
```diff
@@ -5,42 +5,41 @@
...
+     *
+     * @return array<string, JsonSchema>
+     */
+    public function schema(JsonSchema $schema): array
     {
-        return $schema
-            ->string('code')
-            ->description('PHP code to execute (without opening <?php tags)')
-            ->required()
-            ->integer('timeout')
```
ashleyhindle, line 32: We'll want to double check this @nunomaduro - I think there was a way this 'timeout' made it to the ToolExecutor to determine how long the subprocess should run so the timeout was always handle in the subprocess. We don't want to lose this or we'll bring back a bug we just had

src/Mcp/Boost.php
```diff
@@ -10,106 +10,138 @@
...
+     */
+    protected string $version = '0.0.1';
 
-    public string $instructions = 'Laravel ecosystem MCP server offering database schema access, Artisan commands, error logs, Tinker execution, semantic documentation search and more. Boost helps with code generation.';
+    /**
+     * The MCP server's instructions for the LLM.
+     */
+    protected string $instructions = 'Laravel ecosystem MCP server offering database schema access, Artisan commands, error logs, Tinker execution, semantic documentation search and more. Boost helps with code generation.';
 
-    public int $defaultPaginationLength = 50;
```
ashleyhindle, line 19: We'll still want this - the current default is '15' @nunomaduro

src/Console/ExecuteToolCommand.php
nunomaduro: tested this an is working

PR #264, opened 2025-09-18
fix: add csp nonce to browser logger

joostdebruijn:
Boost injects inline Javascript, but this may not work in environments with a CSP-policy. This PR adds the Vite nonce to the injected script tag to make sure it works in CSP-enabled environments (of course, when the CSP-policy allows nonces for scripts).


pushpak1300:
Duplicate of #142 

Thanks.

joostdebruijn:
That one is even nicer, thanks!

PR #277, opened 2025-09-25
[1.x] Adds `boost:update` + allows package authors to publish guidelines

nunomaduro:


ashleyhindle:
It failed on my test project - I guess because I don't have any 3rd party packages with guidelines?

CleanShot 2025-09-27 at 16 22 29@2x

PR #280, opened 2025-09-29
add_hooks_for_third_party_ide

pushpak1300:

### DIFF
--- a/src/Boost.php
+++ b/src/Boost.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Boost;
+
+use Illuminate\Support\Facades\Facade;
+
+/**
+ * @method static void registerCodeEnvironment(string $key, string $className)
+ * @method static array getCodeEnvironments()
+ *
+ * @see \Laravel\Boost\BoostManager
+ */
+class Boost extends Facade
+{
+    protected static function getFacadeAccessor(): string
+    {
+        return BoostManager::class;
+    }
+}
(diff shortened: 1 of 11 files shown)


gldrenthe89:
Great feature. I have created a extension for Augment Code which can be found here:

- Packagist: https://packagist.org/packages/gldrenthe89/laravel-boost-augment-code-extension
- Github: https://github.com/gldrenthe89/laravel-boost-augment-code-extension 

This has been succesfull tested with this proposed PR. Waiting on release 💪

taylorotwell:
I think one thing that's not clear from the readme is what does any of that _actually do_? 🤔 Also, what is a "code environment"?

On this topic, are there other popular coding tools we should just add first-party support for instead?

gldrenthe89:
> I think one thing that's not clear from the readme is what does any of that _actually do_? 🤔 Also, what is a "code environment"?
> 
> On this topic, are there other popular coding tools we should just add first-party support for instead?

I submitted a PR for AugmentCode (https://github.com/laravel/boost/pull/282), but it was rejected in favor of this PR, which I can support. As a result, I created a package for this new feature (https://github.com/gldrenthe89/laravel-boost-augment-code-extension). I’d love to see at least one of these versions made available! ;-)

pushpak1300:
> On this topic, are there other popular coding tools we should just add first-party support for instead?

I think we already support all the major IDEs and AI agents. This change mainly provides a way to extend support for some of the lesser-used ones like Zed, Augment, or Gemini etc, so others can easily add integrations for agents without us having to maintain them as first-party IDEs.

I’ve also updated the `README.md` to better explain what `CodeEnvironment` are, but I’m curious to hear your thoughts. should the class name `CodeEnvironment` be refactored to something like **`Agent`** or **`IDE`** for more clarity?

PR #284, opened 2025-10-01
fix: update copilot instructions path in project detection config

Ale1x:
This pull request updates the location of the Copilot instructions file in the codebase, moving it from `.github/copilot-instructions.md` to `.github/instructions/copilot-instructions.md`. The changes ensure that both the configuration logic and related tests reference the new path.

**Copilot instructions file path update:**

* Updated the `projectDetectionConfig` and `guidelinesPath` methods in `src/Install/CodeEnvironment/Copilot.php` to use the new instructions file location `.github/instructions/copilot-instructions.md` instead of `.github/copilot-instructions.md`. [[1]](diffhunk://#diff-e1c512241441a879be4c5c7cec95dbec9372cf02fa748857ab01634c44dc298eL33-R33) [[2]](diffhunk://#diff-e1c512241441a879be4c5c7cec95dbec9372cf02fa748857ab01634c44dc298eL49-R49)

**Test adjustments:**

* Modified `tests/Unit/Install/CodeEnvironmentsDetectorTest.php` to create and clean up the instructions file in the new `.github/instructions` directory for test setup and teardown.

PR #288, opened 2025-10-05
fix: update copilot instructions generation

akrista:
fix: update Copilot instructions generation to output `.github/instructions/laravel-boost.instructions.md` with frontmatter support

## Summary

This PR is kinda duplicate from #284 and refactors how we generate GitHub Copilot instructions.  
Instead of using `copilot-instructions.md`, we now:

- Mirror the existing `.cursor/rules/laravel-boost.mdc` behavior for consistency using `.github/instructions/laravel-boost.instructions.md`

## Changes

1. **Copilot.php**  
   - Updated the path to reflect the `.github/instructions/…` directory structure and added frontmatter support so you can pass metadata (like `applyTo`) into the generated instructions.

2. **GuidelineWriter.php**  
   - Introduced an `instanceof Copilot` check to route Copilot-specific guideline frontmatter generation appropriately.

3. **CodeEnvironmentsDetectorTest.php**  
   - Updated test paths to reflect the `.github/instructions/…` directory structure.

## Affected files

- `src/CodeEnvironmentsDetectorTest.php`  
- `src/Copilot.php`  
- `src/GuidelineWriter.php`  

Honestly, for GuidelineWriter.php I just added a simple check `instanceof Copilot` using AI, so please let me know if these changes don't fit the contribution guidelines or anything.

## References

- VS Code Copilot custom instructions:  
  https://code.visualstudio.com/docs/copilot/customization/custom-instructions
- GitHub Changelog (2025-07-23):  
  https://github.blog/changelog/2025-07-23-github-copilot-coding-agent-now-supports-instructions-md-custom-instructions/
- GitHub Docs on repository instructions:  
  https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-custom-instructions

### DIFF
--- a/src/Install/CodeEnvironment/Copilot.php
+++ b/src/Install/CodeEnvironment/Copilot.php
@@ -30,7 +30,7 @@ class Copilot extends CodeEnvironment implements Agent
     public function projectDetectionConfig(): array
     {
         return [
-            'files' => ['.github/copilot-instructions.md'],
+            'files' => ['.github/instructions/laravel-boost.instructions.md'],
         ];
     }
 
@@ -46,6 +46,11 @@ class Copilot extends CodeEnvironment implements Agent
 
     public function guidelinesPath(): string
     {
-        return '.github/copilot-instructions.md';
+        return '.github/instructions/laravel-boost.instructions.md';
+    }
+
+    public function frontmatter(): bool
+    {
+        return true;
     }
 }
--- a/src/Install/GuidelineWriter.php
+++ b/src/Install/GuidelineWriter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laravel\Boost\Install;
 
 use Laravel\Boost\Contracts\Agent;
+use Laravel\Boost\Install\CodeEnvironment\Copilot;
 use RuntimeException;
 
 class GuidelineWriter
@@ -60,7 +61,11 @@ class GuidelineWriter
                 // No existing Boost guidelines found, append to end of existing file
                 $frontMatter = '';
                 if ($this->agent->frontmatter() && ! str_contains($content, "\n---\n")) {
-                    $frontMatter = "---\nalwaysApply: true\n---\n";
+                    if ($this->agent instanceof Copilot) {
+                        $frontMatter = "---\napplyTo: \"**\"\n---\n";
+                    } else {
+                        $frontMatter = "---\nalwaysApply: true\n---\n";
+                    }
                 }
 
                 $existingContent = rtrim($content);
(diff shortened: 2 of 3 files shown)


pushpak1300:
> Mirror the existing `.cursor/rules/laravel-boost.mdc` behavior for consistency using `.github/instructions/laravel-boost.instructions.md`.
>
Cursor doesn’t currently support applying rules globally, so having them defined by name makes sense.
In contrast, GitHub Copilot **does** provide a way to apply rules globally through a single file:
> A single [[.github/copilot-instructions.md](https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-a-githubcopilotinstructionsmd-file)](https://code.visualstudio.com/docs/copilot/customization/custom-instructions#_use-a-githubcopilotinstructionsmd-file)
 > Automatically applies to all chat requests in the workspace
 > Stored within the workspace

 Hence, Copilot can handle global rule application natively. 
 
 What are the benefits of using instruction folder vs global rules as in the MR the rules are being applied globally ?

akrista:
> What are the benefits of using instruction folder vs global rules as in the MR the rules are being applied globally ?

none tbh, essentially:

`.github\copilot-instructions.md`
```markdown
<laravel-boost-guidelines>
=== foundation rules ===
...
```

is functionally the same as:

`.github\instructions\laravel-boost.instructions.md`
```markdown
---
applyTo: "**"
---
<laravel-boost-guidelines>
=== foundation rules ===
...
```

> Cursor doesn’t currently support applying rules globally, so having them defined by name makes sense.

[Cursor and other IDEs have used global rules files before](https://web.archive.org/web/20241126183447/https://docs.cursor.com/context/rules-for-ai) (e.g. `.cursorrules`, `.windsurfrules`, `.clinerules`). 

`.cursorrules` is still [supported](https://cursor.com/docs/context/rules#cursorrules), but it will be deprecated. In theory, we could use `.cursorrules` instead of `laravel-boost.mdc`, right?  

However, I don’t think a single file should hold all AI guidelines/rules. While you could group Laravel-specific rules with business or app rules in one global file, each rule set—like Laravel Boost—deserves its own dedicated file rather than being mixed into the global one.

I understand that other AI coding agents like Junie only handles one guideline file (if I’m not mistaken), but I suspect Cursor chose to split rules into multiple folders/files so you can control exactly when and how much context the AI receives.

For example, Windsurf rules only support up to 12,000 characters per file:  
https://docs.windsurf.com/windsurf/cascade/memories#rules-storage-locations

pushpak1300:

That said, I am still thinking about how to give each agent more flexibility about the guidelines it generates, but at this point, I'm not sure if it's right to merge in. Thanks for all the efforts.

PR #294, opened 2025-10-10
Detect frontend package manager

ianrodrigues:
When using bun, pnpm, or yarn instead of npm, Boost's generated guidelines would still reference npm run build and npm run dev, causing confusion.

The guidelines now automatically detect which package manager is in use by checking for lockfiles (bun.lockb, pnpm-lock.yaml, yarn.lock, package-lock.json) and generate the appropriate commands.

This means developers using alternative package managers will see the correct commands in their AI assistant's context without manual intervention.

### DIFF
--- a/.phpactor.json
+++ b/.phpactor.json
@@ -1,4 +1,0 @@
-{
-    "$schema": "/phpactor.schema.json",
-    "language_server_phpstan.enabled": false
-}


ianrodrigues:
I just realized there's a PR for that at #145

PR #301, opened 2025-10-17
Update PHP binary retrieval to use config setting in ToolExecutor

binaryk:
## Fixed
- Update PHP binary retrieval to use config setting in ToolExecutor


pushpak1300:

binaryk:
is that the single explanation @binaryk ? to keep the code size smaller?

PR #306, opened 2025-10-19
Fix casing in directory path for Inertia components

antojdprop:
Fix casing in directory path for Inertia components

### DIFF
--- a/.ai/foundation.blade.php
+++ b/.ai/foundation.blade.php
@@ -12,8 +12,8 @@ This application is a Laravel application and its main Laravel ecosystems packag
 
 @if (! empty(config('boost.purpose')))
 Application purpose: {!! config('boost.purpose') !!}
-@endif
 
+@endif
 ## Conventions
 - You must follow all existing code conventions used in this application. When creating or editing a file, check sibling files for the correct structure, approach, naming.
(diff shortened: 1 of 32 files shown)


.ai/inertia-laravel/core.blade.php
```diff
@@ -1,6 +1,6 @@
 ## Inertia Core
 
-- Inertia.js components should be placed in the `resources/js/Pages` directory unless specified differently in the JS bundler (vite.config.js).
+- Inertia.js components should be placed in the `resources/js/pages` directory unless specified differently in the JS bundler (vite.config.js).
```
pushpak1300, line 3: @antojdprop thanks for the PR.

I thought about this a bit more. Hardcoding the path will break for Laravel 11 starter kits (Jetstream, Breeze, etc). Instead of doing that, can we detect the path dynamically at runtime?

For example, we could use something like:

`preg_match('/[pP]ages/', $appContents, $matches);`

If we detect it in any of the expected files (in priority order), we use that. If we cannot detect it, we fall back to generic wording:

`"Inertia pages are stored in resources/js/ unless specified differently in the JS bundler (vite.config.js)."`

That way it works across versions without assuming folder structure.

PR #307, opened 2025-10-20
Update Readme about .gitignore files

chinmaypurav:
This pull request updates the `README.md` file to include instructions for ignoring local configuration files in version control. This helps prevent accidental commits of user-specific or sensitive configuration.

Documentation improvements:

* Added instructions to add `/.junie` and `.mcp.json` to the `.gitignore` file to prevent committing local configuration files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Next, install the MCP server and coding guidelines:
 php artisan boost:install
 ```
 
-Feel free to add the generated MCP configuration file, guideline files (`.mcp.json`, `CLAUDE.md`, `AGENTS.md`, `junie/`, etc.) and `boost.json` configurationl file to your application's `.gitignore` as these files are automatically re-generated when running `boost:install` and `boost:update`.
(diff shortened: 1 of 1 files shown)


pushpak1300:
This one’s a bit tricky. Some agents support absolute paths, while others need full paths. If you’re using absolute paths, it makes sense to commit the `.mcp.json` file. But if you’re using full paths, committing it doesn’t make sense it really depends on your setup and the agents you’re using.

Since there’s no sensitive info in the config, I’m not convinced this change is needed right now. I’ll hold off merging it until we have a bit more discussion/thoughts around it.

Thanks again for the PR!

chinmaypurav:
> This one’s a bit tricky. Some agents support absolute paths, while others need full paths. If you’re using absolute paths, it makes sense to commit the `.mcp.json` file. But if you’re using full paths, committing it doesn’t make sense it really depends on your setup and the agents you’re using.
> 
> Since there’s no sensitive info in the config, I’m not convinced this change is needed right now. I’ll hold off merging it until we have a bit more discussion/thoughts around it.
> 
> Thanks again for the PR!

There is no sensitive info in the config files, just the paths as per my local environment; so ideally these files should not be committed to git.

As you said, it depends as per individual needs, can we write it like something generic? Like some files do adapt to your local environment, so these files needs to be added to .gitignore? This is more like a caution before the devs commit all touched files. Let me know your thoughts!

pushpak1300:
Okay, I think it makes sense to document the default behaviour suggested by Boost. So here we go.

PR #308, opened 2025-10-20
Change default value for browser_logs_watcher

dragermrb:
Add better default value for browser_logs_watcher.

We typically don't want to inject the watcher on production sites. This way, we disable the watcher by default when the application doesn't have debugging enabled.

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -27,6 +27,6 @@ return [
     | errors within the browser's console to give Boost better context.
     */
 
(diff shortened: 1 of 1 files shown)


pushpak1300:

Unfortunately, I’m going to delay merging this code for now. It’s already configured in the same way, and browser logger isn’t registered if app debug is false. 

https://github.com/laravel/boost/blob/707c7490371e68e1420b27739b33980e08212c62/src/BoostServiceProvider.php#L186

PR #311, opened 2025-10-21
Add DDEV example to README for container usage

devcflynn:
Added DDEV configuration example for using containers.


pushpak1300:

PR #312, opened 2025-10-22
Support flexible frontmatter data structure in Agent contract

GoneTone:
## Description

This PR refactors the `Agent` contract to support flexible frontmatter data structures by replacing the boolean `frontmatter()` method with an array-based `frontMatterData()` method.

### Changes

- Replace `Agent::frontmatter(): bool` with `Agent::frontMatterData(): ?array`
- Update `Cursor` implementation to return `['alwaysApply' => true]` instead of `true`
- Enhance `GuidelineWriter` to handle array-based frontmatter data
- Improve frontmatter detection to support cross-platform line endings (LF, CRLF, CR)
- Update all related tests to reflect the new method signature

### Benefits to End Users

1. **More Flexible Configuration**: Agents can now define multiple frontmatter fields with custom values, not just a boolean flag
2. **Better Extensibility**: Future agents can easily add their own frontmatter fields without changing the contract
3. **Cleaner API**: The new signature `frontMatterData(): ?array` is more explicit about what data is being returned
4. **Cross-Platform Compatibility**: Frontmatter detection now works correctly across different operating systems with varying line ending formats

### Backward Compatibility

This is a **breaking change** that affects the Agent contract. Any custom agent implementations that override the `frontmatter()` method will need to be updated to implement `frontMatterData(): ?array` instead.

**Migration Guide for Custom Agents:**
- If your agent returned `false`, change it to return `null`
- If your agent returned `true`, change it to return an array with your desired frontmatter fields (e.g., `['alwaysApply' => true]`)

### Testing

- All existing tests have been updated to use the new method signature
- Frontmatter-related tests verify proper array handling
- Tests ensure null values are handled correctly for agents that don't require frontmatter

### Example Usage

Before:
```php
public function frontmatter(): bool
{
    return true;
}
```

After:
```php
public function frontMatterData(): ?array
{
    return [
        'alwaysApply' => true,
        // Future fields can be added here
    ];
}
```

This change provides agents with more flexibility to define custom frontmatter configurations tailored to their specific needs.

### DIFF
--- a/src/Install/GuidelineWriter.php
+++ b/src/Install/GuidelineWriter.php
@@ -60,7 +60,7 @@ class GuidelineWriter
                 // No existing Boost guidelines found, append to end of existing file
                 $frontMatter = '';
                 $frontMatterData = $this->agent->frontMatterData();
-                if (! empty($frontMatterData) && ! str_contains($content, "\n---\n")) {
+                if (! empty($frontMatterData) && ! $this->hasFrontMatter($content)) {
                     $frontMatter = "---\n";
                     foreach ($frontMatterData as $key => $value) {
                         $formattedValue = match (true) {
@@ -98,6 +98,13 @@ class GuidelineWriter
         }
 
         return $replaced ? self::REPLACED : self::NEW;
+    }
+
+    protected function hasFrontMatter(string $content): bool
+    {
+        // Check for frontmatter pattern at the start of content
+        // Supports different line endings (LF, CRLF, CR)
+        return (bool) preg_match('/^---[\r\n]+.*?[\r\n]+---[\r\n]/s', $content);
     }
 
     protected function acquireLockWithRetry(mixed $handle, string $filePath, int $maxRetries = 3): void


pushpak1300:
Currently, only the `Cursor` uses `frontmatter`. I’m still exploring how to make the broader code environment more extensible, so it’s unclear if there are immediate benefits to this change yet. Are there any other agents that currently use frontmatter or define it in a different format that would actually benefit from this refactor?

GoneTone:
Based on my knowledge, Windsurf rules use frontmatter.

```
---
trigger: manual
glob:
description:
---
```

`trigger` can be one of the following:

- `manual` - This rule can be manually activated via @mention in Cascade’s input box
- `always_on` - This rule will always be applied
- `model_decision` - Based on a natural language description of the rule the user defines, the model decides whether to apply the rule
- `glob` - Based on the glob pattern that the user defines (e.g. .js, src/**/.ts), this rule will be applied to all files that match the pattern

GoneTone:
We cannot know whether other third-party IDEs / AI Agents support will require custom frontmatter, but I believe this refactoring is meaningful as it allows other third-party IDEs / AI Agents support to customize frontmatter!

pushpak1300:
Thanks for your pull request! 💯 

Unfortunately, I'm going to delay merging this code for now as it introduces a breaking change to the package.

GoneTone:
If it's compatible with the original frontmatter, would you accept this PR?

pushpak1300:
In that case, it makes sense, but I am still on the fence about it as custom implementation can still add it, and I don’t see many benefits to it, but I am really not sure.

GoneTone:
Understood, I won't make any changes for now. Let me know if you need anything!

PR #315, opened 2025-10-24
[1.x] Refactor to use first-class callable

pushpak1300:
Matches rector 2.2.4

similar as https://github.com/laravel/mcp/pull/94


src/Mcp/Tools/DatabaseSchema/MySQLSchemaDriver.php
```diff
@@ -43,7 +43,7 @@ public function getFunctions(): array
     public function getTriggers(?string $table = null): array
     {
         try {
-            if ($table !== null && $table !== '' && $table !== '0') {
+            if (! in_array($table, [null, '', '0'], true)) {
```
joetannenbaum: Doing this in three places, feels like there's a helper method here?

joetannenbaum (changes requested):

PR #329, opened 2025-11-01
Add Sail Support in Guidelines

pushpak1300:
This update introduces first class Sail support across the guidelines.  
When a project installs Boost with Sail enabled, agent instructions now automatically switch to Sail specific commands rather than native CLI commands.

`Example guidelines`

Before  
`- Run the minimum number of tests needed. Use php artisan test with a filename or filter.`

After (Sail selected during `boost:install`)  
`- Run the minimum number of tests needed. Use vendor/bin/sail artisan test with a filename or filter.`

Alongside these dynamic command replacements, this MR adds dedicated Sail usage guidance so LLM agents can reliably work inside Sail environments.

```text
## Laravel Sail

- This project runs inside Laravel Sail's Docker containers. You must run commands through Sail.
- Start services with `vendor/bin/sail up -d` and stop them using `vendor/bin/sail stop`.
- Open the app in a browser with `vendor/bin/sail open`.
- Prefix PHP, Artisan, Composer and Node commands with `vendor/bin/sail`. For example:
  - `vendor/bin/sail artisan migrate`
  - `vendor/bin/sail composer install`
  - `vendor/bin/sail npm run dev`
  - `vendor/bin/sail php script.php`
- View all Sail commands by running `vendor/bin/sail` with no arguments.

### DIFF
--- a/src/Install/GuidelineComposer.php
+++ b/src/Install/GuidelineComposer.php
@@ -227,12 +227,8 @@ class GuidelineComposer
         foreach ($this->packagePriorities as $priorityPackage => $excludedPackages) {
             $packageIsInExclusionList = in_array($package->package()->value, $excludedPackages, true);
 
-            if ($packageIsInExclusionList) {
-                $priorityPackageExists = $this->roster->uses(Packages::from($priorityPackage));
-
-                if ($priorityPackageExists) {
-                    return true;
-                }
+            if ($packageIsInExclusionList && $this->roster->uses(Packages::from($priorityPackage))) {
+                return true;
             }
         }
 

### CODE REVIEW

File: .ai/pint/core.blade.php [hunk 1]
```diff
@@ -1,4 +1,4 @@
 ## Laravel Pint Code Formatter
 
-- You must run `vendor/bin/pint --dirty` before finalizing changes to ensure your code matches the project's expected style.
-- Do not run `vendor/bin/pint --test`, simply run `vendor/bin/pint` to fix any formatting issues.
+- You must run `{{ $assist->bin() }}pint --dirty` before finalizing changes to ensure your code matches the project's expected style.
```
joetannenbaum: Typically, I feel like this would be written as `{{ $assist->bin() }}/pint --dirty`, so that `$assist->bin()` would return the path without the trailing slash for clarity. Thoughts?

src/Install/GuidelineAssist.php
```diff
@@ -24,8 +24,11 @@ class GuidelineAssist
 
     protected static array $classes = [];
 
+    public bool $usesSail;
+
     public function __construct(public Roster $roster)
     {
+        $this->usesSail = (new BoostConfig)->getSail();
```
joetannenbaum: `getSail` returns a boolean? That seems a bit odd. `hasSail`? `usesSail`?

src/Install/GuidelineAssist.php
```diff
@@ -160,11 +159,6 @@ public function enumContents(): string
         return file_get_contents(current($this->enumPaths));
     }
 
-    public function packageGte(Packages $package, string $version): bool
```
pushpak1300, line 163: nit: Unused Method

src/Install/GuidelineConfig.php
```diff
@@ -10,6 +10,8 @@ class GuidelineConfig
 
     public bool $laravelStyle = false;
 
+    public bool $enforceSail = false;
```
joetannenbaum: Is `enforce` the right term here? Feels off.

src/Console/InstallCommand.php
```diff
@@ -530,21 +536,10 @@ protected function shouldUseSail(): bool
...
-    protected function isRunningInsideSail(): bool
-    {
-        return get_current_user() === 'sail' || getenv('LARAVEL_SAIL') === '1';
-    }
-
     protected function buildMcpCommand(McpClient $mcpClient): array
     {
         if ($this->shouldUseSail()) {
-            return ['laravel-boost', './vendor/bin/sail', 'artisan', 'boost:mcp'];
+            return ['laravel-boost', Sail::SAIL_BINARY_PATH, 'artisan', 'boost:mcp'];
```
joetannenbaum: I feel like all Sail related things should be encapsulated in `Sail`, right? So shouldn't this be (pseudo code): `$sail->buildMcCommand()`?

src/Install/GuidelineAssist.php
```diff
@@ -174,4 +168,25 @@ public function nodePackageManager(): string
     {
         return ($this->roster->nodePackageManager() ?? NodePackageManager::NPM)->value;
     }
+
+    public function artisan(): string
+    {
+        return $this->config->enforceSail
+            ? Sail::SAIL_BINARY_PATH.' artisan'
```
joetannenbaum: Again, does it make more sense to have `$sail->artisan()`?

.ai/pint/core.blade.php
pushpak1300: I have refactored this to use `$assist->binCommand('pint')` which feels much cleaner to me overall let me know what do you think.

src/Install/GuidelineAssist.php
pushpak1300: I’m fine with either approach. My intent was to avoid over encapsulation and prevent Sail from being too closely tied to `GuidelineAssist.php`.

I’ve now refactored it to call the static methods from Sail instead. wdyt ?

src/Install/GuidelineConfig.php
pushpak1300: Yes you are right enforce is too strict done refactoring from `enforceSail` -> `usesSail`

wilsenhc:
This is huge, thanks @pushpak1300 🚀🚀

.ai/sail/core.blade.php
```diff
@@ -0,0 +1,14 @@
+@php
+/** @var \Laravel\Boost\Install\GuidelineAssist $assist */
+@endphp
+## Laravel Sail
+
+- This project runs inside Laravel Sail's Docker containers. You MUST execute all commands through Sail.
+- Start services using `{{ $assist->sailBinaryPath() }} up -d` and stop them with `{{ $assist->sailBinaryPath() }} stop`.
+- Open the application in the browser by running `{{ $assist->sailBinaryPath() }} open`.
+- Always prefix PHP, Artisan, Composer, and Node commands** with `{{ $assist->sailBinaryPath() }}`. Examples:
```
hackel, line 9: Hey @pushpak1300 I was just wondering, did you meant to connect up these two asterisks after "commands" to emphasize something?  They stood out to me, looking a bit out of place all by themselves.

pushpak1300: looks typo.

PR #334, opened 2025-11-03
Ignore MCP config update in boost:update

pushpak1300:
Running `boost:update` locally currently rewrites the existing MCP config every time the command runs. This wipes out any manual edits, which interrupts local development workflows. Updating this logic so it respects existing MCP config avoids unexpected resets.

This MR also introduces two new install options to provide more control during setup:

```
--ignore-guidelines   Skip installing AI guidelines  
--ignore-mcp          Skip installing MCP server configuration
```

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -34,7 +34,7 @@ class InstallCommand extends Command
 {
     use Colors;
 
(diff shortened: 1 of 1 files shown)


joetannenbaum:
@pushpak1300 did a refactor that I think is a bit cleaner, let me know what you think.

pushpak1300:
Thanks @joetannenbaum much cleaner 💯 🧼

PR #342, opened 2025-11-05
Update core guidelines to use Laravel's built-in helpers when available

electronic-club:
In many cases, Juni tends to re-implement functionality that already exists in the Laravel framework instead of using the built-in helpers. After a few tests, it seems this guideline helps clarify the preferred approach.


pushpak1300:
Thanks for your pull request! 

PR #348, opened 2025-11-09
add the ability to override 3rd party packages

adrum:
I found it difficult to override 3rd-party package guidelines. This PR aims to simplify this process so it works as expected.


pushpak1300:
In this case, can’t you simply remove the third-party guidelines and then add your own custom guidelines by adding guidelines in the `.ai/` folder ?

emreakay:
@pushpak1300 very nice

adrum:
@pushpak1300 It's not clear to me how to remove the third-party guidelines. I looked through the readme one more time to make sure I wasn't missing anything. 

Can you share instructions on how I could do that? I definitely agree having the ability to provide your own is preferred.

I thought it would be simple to have the ability to override third party ones similar to the built-in guidelines.

pushpak1300:
@adrum you can rerun `php artisan boost:install` select the third party packages you want to include or remove the one which you want to exclude and the guidelines will be updated.

PR #354, opened 2025-11-13
Add note about boost:update command requirements

### LINKED ISSUE #352: Clarify update process for Laravel Boost (no 'boost:update' command)
Problem  
Running php artisan boost:update fails because there is no such command:

Environment  
- PHP 8.4.14 (Herd)  
- Laravel 12.30.1  
- macOS 15.6.1 (arm64)  
- laravel/boost installed v1.2.0; latest available v1.8.0

Expected  
Either a dedicated update command (or alias), or docs clearly stating the update path.

Actual  
No update command exists; users may be unsure about the correct update steps.

Request  
- Please clarify in the documentation that the user has to update Laravel Boost to the latest version to be able to update

bjoernzosel:
This PR adds a helpful note in the "Keeping Guidelines Up-to-Date" section to clarify the requirements for the `boost:update` command.

   ## Changes

   - Added a note block explaining that users need to be on the latest Laravel Boost version
   - Clarified that Laravel Boost only registers when `BOOST_ENABLED=true` or `APP_DEBUG=true` is set in `.env`

   ## Why

   This helps users troubleshoot issues when the `boost:update` command is not appearing in their artisan command list.

   Fixes #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,9 @@ You may also automate this process by adding it to your Composer "post-update-cm
 }
 ```
 
+> [!NOTE]
+> If you do not see a `boost:update` command, make sure you are on the latest Laravel Boost version. Laravel Boost will only be registered if `BOOST_ENABLED=true` is set in your `.env` file (or if `APP_DEBUG=true`).
+
 ## Available Documentation
 
 | Package | Versions Supported |


pushpak1300:
Thanks for your pull request!

Issue #358, opened 2025-11-18
Bug Report: ResponseFactory::map() Method Not Found in CallToolWithExecutor


RedArchon:
### Laravel Package Version

1.8.0

### Laravel Version

11.46.1

### PHP Version

8.3.27

### System Info

Docker container

### Description

# Bug Report: ResponseFactory::map() Method Not Found in CallToolWithExecutor

## Summary

When executing MCP tools via Laravel Boost, the server crashes with `BadMethodCallException: Method Laravel\Mcp\ResponseFactory::map does not exist`. The issue occurs in `CallToolWithExecutor.php` where the code attempts to call `->map()` directly on a `ResponseFactory` object instead of first calling `->responses()` to get the underlying Collection.

## Environment

- **Laravel Boost Version**: 1.8.0
- **Laravel MCP Version**: 0.3.4
- **Laravel Framework Version**: 11.46.1
- **PHP Version**: 8.3.27
- **Environment**: Docker container (though issue likely affects all environments)

## Steps to Reproduce

1. Install Laravel Boost: `php artisan boost:install`
2. Configure MCP server in `~/.cursor/mcp.json` (or similar IDE configuration)
3. Start the MCP server: `php artisan boost:mcp`
4. Connect to the MCP server (connection succeeds, tools are discovered)
5. Attempt to execute any MCP tool (e.g., `get-config`, `application-info`)
6. Server crashes with the error below

### Error Message

```
BadMethodCallException: Method Laravel\Mcp\ResponseFactory::map does not exist.
```

### Stack Trace

```
at vendor/laravel/framework/src/Illuminate/Macroable/Traits/Macroable.php:115
#0 vendor/laravel/boost/src/Mcp/Methods/CallToolWithExecutor.php(62): Laravel\Mcp\ResponseFactory->__call('map', Array)
#1 vendor/laravel/mcp/src/Server/Methods/Concerns/InteractsWithResponses.php(37): Laravel\Boost\Mcp\Methods\CallToolWithExecutor->{closure}()
#2 vendor/laravel/boost/src/Mcp/Methods/CallToolWithExecutor.php(61): Laravel\Boost\Mcp\Methods\CallToolWithExecutor->toJsonRpcResponse()
#3 vendor/laravel/mcp/src/Server.php(263): Laravel\Boost\Mcp\Methods\CallToolWithExecutor->handle()
```

### Log Output

When the error occurs, PHP error output is sent to stdout as plain text, breaking the JSON-RPC protocol:

```
[...]

PR #359, opened 2025-11-18
Fix CallToolWithExecutor Error

RedArchon:
# Fix ResponseFactory::map() Method Not Found Error in CallToolWithExecutor
For issue: https://github.com/laravel/boost/issues/358

## Description

This PR fixes a critical bug in `CallToolWithExecutor` where the code attempts to call `->map()` directly on a `ResponseFactory` object, causing a `BadMethodCallException` when executing MCP tools. The fix ensures that `->responses()` is called first to get the underlying Collection before using `->map()` and `->contains()`.

## Problem

When executing MCP tools via Laravel Boost, the server crashes with:
```
BadMethodCallException: Method Laravel\Mcp\ResponseFactory::map does not exist.
```

This occurs because `ResponseFactory` doesn't have a `map()` method. According to the `ResponseFactory` API in `laravel/mcp`, you must first call `->responses()` to get the underlying `Collection`, then call `->map()` on that collection.

### Root Cause

In `src/Mcp/Methods/CallToolWithExecutor.php` at line 62, the code was calling `->map()` directly on a `ResponseFactory` object:

```php
// Before (broken)
return $this->toJsonRpcResponse($request, $response, fn ($responses): array => [
    'content' => $responses->map(fn ($response) => $response->content()->toTool($tool))->all(),
    'isError' => $responses->contains(fn ($response) => $response->isError()),
]);
```

However, the callback receives a `ResponseFactory` object (not a Collection), which doesn't have `map()` or `contains()` methods.

## Solution

The fix updates the callback to call `->responses()` first to get the Collection, then call `->map()` and `->contains()` on that collection:

```php
// After (fixed)
return $this->toJsonRpcResponse($request, $response, fn ($responseFactory): array => [
    'content' => $responseFactory->responses()->map(fn ($response) => $response->content()->toTool($tool))->all(),
    'isError' => $responseFactory->responses()->contains(fn ($response) => $response->isError()),
]);
```

## Changes Made

- **File**: `src/Mcp/Methods/CallToolWithExecutor.php`
  - Updated the callback parameter name from `$responses` to `$responseFactory` to better reflect the actual type
  - Added `->responses()` calls before `->map()` and `->contains()` to access the underlying Collection

## Benefit to End Users

- **Fixes MCP tool execution**: Users can now successfully execute MCP tools (e.g., `get-config`, `application-info`, `tinker`) without the server crashing
- **Prevents connection failures**: The MCP server no longer crashes and closes connections when tools are executed
- **Improves reliability**: The MCP server remains stable and functional for tool execution
- **Better error handling**: Tool execution errors are now properly handled and returned as JSON-RPC responses instead of causing server crashes

## Why This Doesn't Break Existing Features

- **No API changes**: The fix only corrects the internal implementation to use the correct `ResponseFactory` API
- **Backward compatible**: The method signature and return types remain unchanged
- **Same behavior**: The functionality remains the same, just using the correct API calls
- **No breaking changes**: All existing code that uses `CallToolWithExecutor` will continue to work, but now correctly

## How This Makes Building Web Applications Easier

- **Enables AI-assisted development**: Laravel Boost's MCP integration allows AI assistants (like Cursor, Claude, etc.) to interact with Laravel applications, making development faster and more efficient
- **Improves developer experience**: Developers can now use MCP tools to query configuration, execute tinker commands, check routes, and more without manual intervention
- **Reduces debugging time**: When MCP tools work correctly, developers can quickly get information about their application state through AI assistants
- **Enhances productivity**: The ability to execute tools programmatically through MCP enables more sophisticated AI-assisted workflows

## Related Issues

This fixes the bug described in the issue where MCP tools fail to execute with `BadMethodCallException: Method Laravel\Mcp\ResponseFactory::map does not exist.`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "illuminate/contracts": "^10.49.0|^11.45.3|^12.28.1",
         "illuminate/routing": "^10.49.0|^11.45.3|^12.28.1",
         "illuminate/support": "^10.49.0|^11.45.3|^12.28.1",
-        "laravel/mcp": "^0.3.2",
+        "laravel/mcp": "^0.3.4",
         "laravel/prompts": "0.1.25|^0.3.6",
         "laravel/roster": "^0.2.9"
     },


RedArchon:
@pushpak1300 should I go ahead and remove the "Testing Instructions" section of the PR description?

PR #360, opened 2025-11-18
Add Gemini

iruoy:
This PR adds support for Gemini which is the last missing entry of the big players in my eyes.

Gemini has more than caught up now and offers significant advantages even.


pushpak1300:
Thanks. I’m not fully convinced Gemini belongs in the core yet. Let’s give it a little time and see if it actually good enough to add it in Boost itself as every IDE we add we have to support/maintain them in the package.

In the meantime you can still ship your package using the guidelines here https://github.com/laravel/boost?tab=readme-ov-file#adding-support-for-other-ides--ai-agents so you can use it in your local workflow.

PR #361, opened 2025-11-19
tests: adds missing opencode in tests

MrPunyapal:
This pull request makes a minor update to the `BoostManagerTest` unit test by adding the `OpenCode`.


pushpak1300:
Thanks !

PR #364, opened 2025-11-20
Extend Codex functionality with MCP config

### LINKED ISSUE #363: Codex MCP server config
Codex currently does not have a way to define MCP servers per project directory. However it is possible to define relative paths within the global codex config file which lives in the users home directory. 

Putting this here to start a discussion on boost being able to add the config to the users global codex config.

Image

pushpak1300:
Resolves #363.

We use the MCP shell installation strategy to add MCP config support as described in the Codex [MCP CLI](https://developers.openai.com/codex/mcp/#configuration---cli) docs. This enables Codex to install MCP servers directly via CLI. Although [Codex](https://github.com/openai/codex/issues/2628) does not support project specific config yet, it now supports relative paths in the MCP config, which we leverage to add support for Laravel Boost.

Tested this with sail, herd and it work without any issues. 

Screenshot 2025-11-20 at 4 14 35 PM
Screenshot 2025-11-20 at 4 19 07 PM
Screenshot 2025-11-20 at 4 29 06 PM

### DIFF
--- a/src/Install/CodeEnvironment/Codex.php
+++ b/src/Install/CodeEnvironment/Codex.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Laravel\Boost\Install\CodeEnvironment;
 
-use Illuminate\Support\Facades\Process;
 use Laravel\Boost\Contracts\Agent;
 use Laravel\Boost\Contracts\McpClient;
 use Laravel\Boost\Install\Enums\McpInstallationStrategy;


jsandfordhughescoop:
@pushpak1300 - The only issue with it is that it sets the herd `SITE_PATH` as an absolute path however in my testing by setting it to `*` codex uses the relative path it was executed in.

pushpak1300:
Thanks @jsandfordhughescoop for pointing that out. As [Herd](https://herd.laravel.com/docs/macos/advanced-usage/ai-integrations#manual-configuration) support global level mcp without env variable i've removed env key.

> This is the standard configuration that works in most MCP clients. Make sure to replace YOUR-SITE-PATH with the base path to your Laravel project or remove the env key with the site path if you’re using the global installation.

PR #387, opened 2025-12-11
Mention when private constructors are ok

cosmastech:
Happened to see this and thought I would submit a patch. Might not even be particularly necessary to add, but figured I'd offer it up.


pushpak1300:
I'll let @taylorotwell decide this if we need this in guideline or not.

pushpak1300 (approved):
I definitely think we should improve this type of guidance which are paper cut and easy to fix.

PR #389, opened 2025-12-15
Add Resource and Prompts for Package Guidelines

pushpak1300:
#### Context Management
Previously, all third-party package guidelines would need to be added to the global context, which could grow unboundedly as more packages are integrated. 

The current approach:

- **Reduces global context bloat**: Only guidelines for actively used packages are loaded
- **Improves token efficiency**: Each AI interaction carries smaller context, leaving room for more application-specific information

#### Example
Consider a package that's only used in the admin dashboard. Adding its guidelines to the global context means every single LLM request across the entire application carries this information even when working on features unrelated to that package. With this feature, the package guidelines are only available as a Resource/Prompt that can be accessed when needed.

#### How It Works

1. **Package Discovery**: The system automatically discovers packages with Boost guidelines using `Composer::packagesDirectoriesWithBoostGuidelines()`

2. **Dynamic Registration**: For each package containing a `core.blade.php` file:
   - A `PackageGuidelineResource` instance is created (makes guidelines available as a searchable resource)
   - A `PackageGuidelinePrompt` instance is created (makes guidelines available as context when explicitly requested)

3. Guidelines are included:
   - As **Prompts** - allowing the LLM to request them when working with that package
   - As **Resources** - allowing users to explicitly fetch and review the guidelines

#### Demo
- https://www.loom.com/share/e25f8634eae84995b397dd39a153e834

Reference
- https://code.claude.com/docs/en/mcp#use-mcp-resources
- https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_use-mcp-prompts
- https://cursor.com/docs/context/mcp#protocol-support

### DIFF
--- a/src/Mcp/Boost.php
+++ b/src/Mcp/Boost.php
@@ -113,12 +113,12 @@ class Boost extends Server
      */
     protected function discoverResources(): array
     {
-        return $this->filterPrimitives(
-            [
-                Resources\ApplicationInfo::class,
-                ...$this->discoverThirdPartyPrimitives(Resource::class)],
-            'resources'
-        );
+        $availableResources = [
+            Resources\ApplicationInfo::class,
+            ...$this->discoverThirdPartyPrimitives(Resource::class),
+        ];
+
+        return $this->filterPrimitives($availableResources, 'resources');
     }
 
     /**


src/Mcp/Boost.php
```diff
@@ -125,16 +113,65 @@ protected function discoverTools(): array
      */
     protected function discoverResources(): array
     {
-        return $this->discoverPrimitives([
-            Resources\ApplicationInfo::class,
-        ], 'resources');
+        return $this->filterPrimitives(
+            [
```
joetannenbaum: Can we fix the formatting of the array braces here?

PR #392, opened 2025-12-17
Allow configurable guideline paths for AI agents

jordanpartridge:
## Problem

When using Boost in monorepos or projects with specific documentation structures, the default guideline file locations (e.g., `CLAUDE.md` in project root) may not fit the project's organization. Teams may want to consolidate AI guidelines in a dedicated directory or point agents to a shared documentation location.

## Solution

This PR adds the ability to customize where Boost writes guideline files for each code environment via Laravel's configuration system. Each agent can have its path configured independently through the `boost.code_environments.*` config namespace.

### Usage

Users can configure custom guideline paths by publishing the Boost config file and adding a `code_environments` section:

```php
// config/boost.php
return [
    // ... existing config

    'code_environments' => [
        'claude_code' => [
            'guidelines_path' => 'docs/ai/CLAUDE.md',
        ],
        'cursor' => [
            'guidelines_path' => 'docs/ai/cursor-rules.mdc',
        ],
        // ... configure other environments as needed
    ],
];
```

Or set it programmatically in a service provider:

```php
config(['boost.code_environments.claude_code.guidelines_path' => 'docs/ai/CLAUDE.md']);
```

Then run `php artisan boost:update` to regenerate guidelines at the new paths.

### Default Paths

If not configured, each environment uses its standard default:

| Environment | Default Path |
|------------|--------------|
| Claude Code | `CLAUDE.md` |
| Codex | `AGENTS.md` |
| Copilot | `.github/copilot-instructions.md` |
| Cursor | `.cursor/rules/laravel-boost.mdc` |
| Gemini | `GEMINI.md` |
| OpenCode | `AGENTS.md` |
| PhpStorm | `.junie/guidelines.md` |

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 return [
+
     /*
     |--------------------------------------------------------------------------
     | Boost Master Switch
@@ -27,4 +28,5 @@ return [
     */
 
     'browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
+
 ];


pushpak1300:
I don’t think this change will be used by many people, so I think it should not be added as a default configuration in config `boost.php`. For now, it can be modified manually if needed. I also don’t want to document this feature at this stage. I’ve updated the MR description accordingly.

jordanpartridge:
> I don’t think this change will be used by many people, so I think it should be added as a default configuration in config `boost.php`. For now, it can be modified manually if needed. I also don’t want to document this feature at this stage. I’ve updated the MR description accordingly.

That makes sense I want quite sure about the change but I figured it would be very useful to a few people.

PR #403, opened 2025-12-27
Add `--compact` flag to testing examples

ohnotnow:
The compact printer can save a lot of tokens being passed to the model.

Quite often the model will run a test file or whole suite and receive 100s or 1000s of lines of :

```
✓ it injects script in html responses with dataset "with head and body tags"
✓ executes code with different return types with dataset "float"
```

This can burn through a _lot_ of tokens.  By telling it to use the `--compact` flag the output is just a series of `.`'s - but the model still gets the full output for tests which fail.

I've been using this instruction in my local .ai/ guidance and it seems to help quite a bit - especially if the model is in a bit of a 'rut' running tests over and over.

### DIFF
--- a/.ai/enforce-tests.blade.php
+++ b/.ai/enforce-tests.blade.php
@@ -4,4 +4,4 @@
 ## Test Enforcement
 
 - Every change must be programmatically tested. Write a new test or update an existing test, then run the affected tests to make sure they pass.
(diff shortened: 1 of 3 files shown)


pushpak1300 (approved):
Thanks. 

Going forward we somehow have to figure it out in tools only and filter them automatically if they are running in agent context.

joetannenbaum (approved):

Issue #336, opened 2025-11-04
Agent Skills Support


Vasiliy-Bondarenko:
Hey, did you consider using Calude Code skills instead of dropping all instructions into single .md file?.. That would save a lot of context window space, keep the mode more focused. Right now in most sessions i barely use 1% of all boost instructions. So it would be nice if most of them live in separate skills.

Personally i have moved quite a few CLAUDE.md sections and custom commands into skills and it works great as long as i can tune a trigger properly.

pushpak1300:
This on my list to explore. Thanks

Vasiliy-Bondarenko:
great! can't wait to try it out!

gregholland:
Would love to see this out of the box too.

brigleb:
This would be amazing. On my projects, Boost uses up effectively 20% of my available context window because all of the instructions are always loaded. It seems like skills would be an excellent solution for this!

nikspyratos:
Codex is also starting to [support skills](https://github.com/openai/codex/blob/main/docs/skills.md#skills-experimental), so if this does become a wider standard across agents it'll be useful to implement.

Otherwise, for now I suppose the ideal implementation would be to split out Boost's instructions to skill files *only* when you're using Claude (or experimental use in Codex) and no other agent (that isn't yet Skills-capable).

pushpak1300:
Opencode is on the way i just hope this become standard soon so we can prioritise this, https://github.com/sst/opencode/issues/3235

pushpak1300:
LFG https://agentskills.io/home#adoption

skattabrain:
Yep, I'm here looking for the same. Skills is where it's at, keep the context neat. This will take boost to the next level!

nick-potts:
Added a simple Skills support #407

keybrdist:
I've gotten this working by disabling laravel-boost at the project level, and then defining it in skills so that boost MCP is only summoned when it is needed.  It works nicely, but still testing mileage and token usage differences.

<img width="852" height="617" alt="Image" src="https://github.com/user-attachments/assets/3542fe5e-28bb-4107-b63a-bc975e1ee0e2" />

`./.opencode/skill/laravel-boost/SKILL.md`

```
---
name: laravel-boost
description: Laravel development tools - tinker, database inspection, routes, config, debugging. Load when working on Laravel backend tasks.
mcp:
  laravel-boost:
    command: php
    args: ["artisan", "boost:mcp"]
---

# Laravel Boost Tools

[...]

PR #407, opened 2026-01-04
Skills

nick-potts:
#336


nick-potts:
Don't think skills are quite ready for prime time. Its extremely tough to get them to activate.

pushpak1300:
Thank you for this, We’re currently conducting experiments to determine if skills are as effective as usual guidelines.  Therefore, I’ll close this until it's finalised.

keybrdist:
@nick-potts curious if you can share some metric. How much reduction in token usage did you see you attempted to rely on just the skills?

PR #419, opened 2026-01-12
Add support for configuring custom executable paths

oddvalue:
Allow users to configure custom paths for php, artisan, composer, and other executables via `boost:install --configure-executables`. This enables support for Docker environments, custom PHP installations (e.g., Herd Pro multi-version), and other non-standard
   setups where default executable paths don't work.

Custom executable configuration is stored in boost.json and automatically overrides Laravel Sail wrappers when set. Configuration includes php, artisan, composer, vendor_bin directory, and node package manager paths.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -617,7 +617,7 @@ class InstallCommand extends Command
 
         // Only prompt for node if different from detected
         $roster = app(\Laravel\Roster\Roster::class);
-        $detectedManager = $roster->nodePackageManager()?->value ?? 'npm';
+        $detectedManager = $roster->nodePackageManager()->value ?? 'npm';
         $configureNode = confirm(
             label: "Configure custom Node package manager? (detected: {$detectedManager})",
             default: false,
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -208,7 +208,7 @@ class GuidelineAssist
     public function composerCommand(string $command): string
     {
         // Priority: Custom config > Sail > Default
-        if ($executables = $this->config->executables) {
+        if (($executables = $this->config->executables) instanceof ExecutableConfig) {
             $composer = $executables->composer;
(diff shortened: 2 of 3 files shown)


james2doyle:
I think the term "path" being used here is incorrect. As these are really the "commands" to use for each tool.

The question should be, "PHP command? Answer: `herd php`"

Given the current prompt, "PHP executable path? Answer: `/Users/myusername/Library/Application Support/Herd/bin/php83`" would be the correct answer for what the PHP path is.

Personally, I believe that Boost should detect Herd (just like it does for sail) and it should be set up automatically.

From what I can tell here, I would need to set my custom "path" for php (herd php), artisan (herd php artisan), and composer (herd composer), for herd to work as expected. And that is not including node. This seems like a lot to configure

oddvalue:
> I think the term "path" being used here is incorrect. As these are really the "commands" to use for each tool.
> 
> The question should be, "PHP command? Answer: `herd php`"
> 
> Given the current prompt, "PHP executable path? Answer: `/Users/myusername/Library/Application Support/Herd/bin/php83`" would be the correct answer for what the PHP path is.
> 
> Personally, I believe that Boost should detect Herd (just like it does for sail) and it should be set up automatically.
> 
> From what I can tell here, I would need to set my custom "path" for php (herd php), artisan (herd php artisan), and composer (herd composer), for herd to work as expected. And that is not including node. This seems like a lot to configure

I agree that Herd should be detected automatically, given that it's a first-party tool. This would at least allow users to get things up and running with it until that's a feature.

I've never actually used Herd or Valet. Is there a common way to detect that you are in a Herd environment when using it? Like the `LARAVEL_SAIL: 1` env variable when using Sail.

pushpak1300:
A simpler and cleaner solution would be to define a single variable in the configuration. If someone needs to override it, they can simply update the PHP path in `boost.php`. This approach would solve multiple problems without introducing the complexity of current implementation.

pushpak1300:
Closing this in favour of #437

PR #427, opened 2026-01-17
Added Windsurf IDE Support

Frostist:
I have added the code that will allow Windsurf to be added as an IDE with Laravel Boost.

Windsurf currently as of writing does not have MCP support for local repos only system wide, which is fine, but the document that is added to the local repo would need to be manually copied each time unless the user installs Laravel Boost.


pushpak1300:
Thanks for your pull request!

Frostist:
Thats understandable! Thanks man

pushpak1300:
https://github.com/GoneTone/laravel-boost-windsurf-extension

PR #430, opened 2026-01-18
Respect platform PHP version from composer.json

SanderMuller:
This PR fixes an issue where Boost uses the system PHP version instead of the project's configured platform PHP version.

### Problem

When running `php artisan boost:update`, Boost uses the system's PHP binary (e.g., `php8.5`) even when `composer.json` specifies a different platform version (e.g., `8.4`).

**Example:**
- System PHP: 8.5
- `composer.json`: `"config.platform.php": "8.4.0"`
- Current: MCP uses `php8.5`
- Expected: MCP should use `php8.4`

### Solution

- Added `Composer::platformPhpVersion()` to read platform PHP from `composer.json`
- Updated `CodeEnvironment::getPhpPath()` to respect this configuration
- Falls back to system PHP when no platform version is configured

### Changes

- New method: `Composer::platformPhpVersion()`
- Updated: `CodeEnvironment::getPhpPath()` to check platform PHP first
- Added tests for all scenarios (three-part version, two-part version, forced absolute paths)

### DIFF
--- a/src/Install/CodeEnvironment/CodeEnvironment.php
+++ b/src/Install/CodeEnvironment/CodeEnvironment.php
@@ -12,6 +12,7 @@ use Laravel\Boost\Install\Detection\DetectionStrategyFactory;
 use Laravel\Boost\Install\Enums\McpInstallationStrategy;
 use Laravel\Boost\Install\Enums\Platform;
 use Laravel\Boost\Install\Mcp\FileWriter;
+use Laravel\Boost\Support\Composer;
 
 abstract class CodeEnvironment
 {
@@ -40,6 +41,16 @@ abstract class CodeEnvironment
 
     public function getPhpPath(bool $forceAbsolutePath = false): string
     {
+        $platformPhpVersion = Composer::platformPhpVersion();
+
+        if ($platformPhpVersion !== null && ! $this->useAbsolutePathForMcp() && ! $forceAbsolutePath) {
+            $versionParts = explode('.', $platformPhpVersion);
+
+            if (count($versionParts) >= 2) {
+                return 'php'.$versionParts[0].'.'.$versionParts[1];
+            }
+        }
+
         return ($this->useAbsolutePathForMcp() || $forceAbsolutePath) ? PHP_BINARY : 'php';
     }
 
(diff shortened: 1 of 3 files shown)


pushpak1300:
A simpler and cleaner solution would be to define a single variable in the configuration. If someone needs to override it, they can simply update the PHP path in` boost.php`. This approach would solve multiple problems without introducing the complexity of current implementation. Also I feel which solves #419 directly without overcomplicating stuff.

SanderMuller:
> A simpler and cleaner solution would be to define a single variable in the configuration. If someone needs to override it, they can simply update the PHP path in` boost.php`. This approach would solve multiple problems without introducing the complexity of current implementation. Also I feel which solves #419 directly without overcomplicating stuff.

I agree that this has some complexity, but your suggestion would require maintaining a php version in a config which is already specific according to the Composer spec in `composer.json`

agarzon-orangerine:
> This PR fixes an issue where Boost uses the system PHP version instead of the project's configured platform PHP version.
> 
> ### Problem
> When running `php artisan boost:update`, Boost uses the system's PHP binary (e.g., `php8.5`) even when `composer.json` specifies a different platform version (e.g., `8.4`).
> 
> **Example:**
> 
> * System PHP: 8.5
> * `composer.json`: `"config.platform.php": "8.4.0"`
> * Current: MCP uses `php8.5`
> * Expected: MCP should use `php8.4`
> 
> ### Solution
> * Added `Composer::platformPhpVersion()` to read platform PHP from `composer.json`
> * Updated `CodeEnvironment::getPhpPath()` to respect this configuration
> * Falls back to system PHP when no platform version is configured
> 
> ### Changes
> * New method: `Composer::platformPhpVersion()`
> * Updated: `CodeEnvironment::getPhpPath()` to check platform PHP first
> * Added tests for all scenarios (three-part version, two-part version, forced absolute paths)

I would actually expect the opposite. I rather prefer to work on the version that my system is actually using vs the one that supposed to use.

For example, my composer says 8.2, but my system is using 8.4 and of course I would prefer everything to works towards the real environmet.

That is just my oppinion.

pushpak1300:
Thanks for your pull request!

PR #435, opened 2026-01-19
Feature | Warn existing installation overwrite on install

Nicholasbell03:
## Summary                                                                                                                                                             

This PR adds a warning prompt when running boost:install on a project that already has Boost installed. It informs users about the boost:update command and asks for confirmation before proceeding with a fresh install that would overwrite existing configuration.                                                                   
                                                                                                                                                                      
## Motivation                                                                                                                                                          
                                                                                                                                                                      
When running boost:install on a project where Boost is already configured, the command silently overwrites the existing boost.json configuration without warning. Users who are unaware that boost:update exists (as I was) may inadvertently lose their carefully configured settings.                                               
                                                                                                                                                                      
This change improves the developer experience by:                                                                                                                   
  - Alerting users that Boost is already installed                                                                                                                    
  - Educating users about the boost:update command as the recommended approach for existing installations                                                             
  - Requiring explicit confirmation before overwriting existing configuration                                                                                         
                                                                                                                                                                      
## Changes                                                                                                                                                             
                                                                                                                                                                      
  - Added exists() method to Config class to detect existing installations                                                                                            
  - Added --force flag to boost:install to bypass the confirmation prompt                                                                                             
  - Added confirmation prompt when boost.json exists (respects --no-interaction)                                                                                      
  - Added tests for both the new Config::exists() method and the install command behavior
  - Update readme                                                                             
                                                                                                                                                                      
## Usage                                                                                                                                                               
                                                                                                                                                                      
### Shows warning and prompts for confirmation if boost.json exists                                                                                                   
  php artisan boost:install

<img width="1495" height="413" alt="image" src="https://github.com/user-attachments/assets/6565faab-35f2-47d8-b423-91f0a91327c2" />

                                                                                                                                                                      
### Bypasses the check (for automation/scripting)                                                                                                                     
  php artisan boost:install --force                                                                                                                                   
                                                                                                                                                                      
The check is automatically skipped when using --no-interaction, ensuring boost:update continues to work as expected.

### DIFF
--- a/tests/Feature/Console/InstallCommandExistingInstallTest.php
+++ b/tests/Feature/Console/InstallCommandExistingInstallTest.php
@@ -16,16 +16,16 @@ afterEach(function (): void {
 test('install command skips existing installation check with --no-interaction flag', function (): void {
     $this->artisan('boost:install', ['--no-interaction' => true, '--ignore-mcp' => true])
         ->assertSuccessful();
-});
+})->skipOnWindows();
 
 test('install command skips existing installation check with --force flag', function (): void {
     $this->artisan('boost:install', ['--force' => true, '--no-interaction' => true, '--ignore-mcp' => true])
         ->assertSuccessful();
-});
+})->skipOnWindows();
 
 test('install command proceeds when no existing installation exists', function (): void {
     (new Config)->flush();
 
     $this->artisan('boost:install', ['--no-interaction' => true, '--ignore-mcp' => true])
         ->assertSuccessful();
-});
+})->skipOnWindows();


pushpak1300:
Hey, instead of adding a separate prompt (which I think is little bad for UX), I’ve added a warning in the new MR #439:
https://github.com/laravel/boost/pull/439/changes#diff-5cffa5eda5bfb718c5f416e0c6e95cf6913e8a092d70581ec5f3e1dfae349941R228

Screenshot 2026-01-20 at 8 01 06 PM

I’ll defer merging this for now, but thanks for the pull request.

Issue #429, opened 2026-01-18
Introduced Sail support breaks development within devcontainer


smares:
### Boost Version

1.8.10

### Laravel Version

12.46.0

### Agent/IDE Name & Version

VS Code 1.108.1

### PHP Version

8.4.16

### System Info

Ubuntu 25.10 with Laravel Sail but developing within devcontainer

### Description

The introduced Sail support that always uses `sail` to launch the MCP server and also instructs the agent to launch commands via `sail` breaks development within devcontainers where everything is supposed to be executed without `sail`.

### Steps To Reproduce

1. Set up Laravel application with Sail and choose to use devcontainers
2. Launch VS Code and point to the newly created application and when asked, choose to reopen the project inside a devcontainer
3. Attempt to launch the MCP server

smares:
As a side-note, it makes no difference if I install boost on the host via `sail` commands or within the devcontainer.

pushpak1300:
I haven’t used Dev Containers before, but I tried reproducing the issue locally and was able to see it. This seems to happen only in VS Code when running inside a Dev Container.

When using VS Code Copilot with a Dev Container, we need to run php artisan commands instead of vendor/bin/sail, since Sail is already handled by Docker itself.

I have a couple of questions to clarify. First, how does the Dev Container workflow usually work? As far as I understand, claude-code environments can still run everything from outside the container, and only VS Code requires commands like `php artisan` to be configured differently.

Second, is there a reliable way to detect whether a command is being executed inside a Dev Container or not? I’m not fully sure how to handle this case yet.

I’d really appreciate your guidance on this. Thank you so much.

smares:
Would it be an option to give the user the choice?

Edit: There are some environment variables you might be able to rely on:
`REMOTE_CONTAINERS=true`, presence of `VSCODE_IPC_HOOK_CLI` or `REMOTE_CONTAINERS_IPC`

pushpak1300:
> Would it be an option to give the user the choice?

I really don't want to add more options but i think we can show this option only if the `devcontainers` file is exists and thats too only for the vscode. 🤔 

I will check how the UX will be based on this `REMOTE_CONTAINERS=true`

pushpak1300:
Can you overwrite this env variables and make it work ? This will take preference over all the config. 

```
BOOST_PHP_EXECUTABLE_PATH=php
BOOST_COMPOSER_EXECUTABLE_PATH=composer
[...]

PR #437, opened 2026-01-20
[1.x] Add configurable PHP and Composer paths

aivis:
## Summary

This PR adds support for configurable PHP and Composer binary paths through two new configuration options:
- `boost.default_php_bin` - Configure the PHP binary path (defaults to `php`)
- `boost.default_composer_bin` - Configure the Composer binary path (defaults to `composer`)
- It does not change the existing behaviour in any way unless the new config keys are used

## Motivation

Many developers use containerized or specialized development environments like **Lando**, **DDEV**, **Nix**, **Homebrew**, or custom Docker setups where PHP and Composer binaries are not available at the standard `php` and `composer` paths:
- Lando commands need to be prefixed with `lando` (`lando composer` etc.).
- Running multiple PHP versions side-by-side (`/usr/local/bin/php8.3`) 
- Using Homebrew-installed PHP with non-standard paths
- Working in environments where the system PATH doesn't include the desired PHP/Composer binaries

This feature allows developers to configure their environment and have Boost generate correct commands for their setup.
I did not add these config keys to the `config/boost.php` file as it seems to be the case for other custom config keys. 

## Changes

- **`CodeEnvironment::getPhpPath()`** - Now reads from `config('boost.default_php_bin', 'php')` when not forcing absolute paths
- **`GuidelineAssist::composerCommand()`** - Now reads from `config('boost.default_composer_bin', 'composer')` when not using Sail
- **`GuidelineAssist::artisan()`** - Now uses the configured PHP binary for artisan commands when not using Sail
- Adds tests matching existing patterns to cover the introduced changes

## Usage

Add the following to your `config/boost.php` file:

```php
return [
    // ... existing config

     'commands' => [                                                                                                                                                                                                         
         'php' => env('BOOST_PHP_BIN', 'php'),                                                                                                                                                                               
         'composer' => env('BOOST_COMPOSER_BIN', 'composer'),                                                                                                                                                                
     ],      
];
```

### DIFF
--- a/src/Install/CodeEnvironment/CodeEnvironment.php
+++ b/src/Install/CodeEnvironment/CodeEnvironment.php
@@ -40,9 +40,9 @@ abstract class CodeEnvironment
 
     public function getPhpPath(bool $forceAbsolutePath = false): string
     {
-        $phpBinaryPath = config('boost.commands.php_binary', 'php');
+        $phpBinaryPath = config('boost.executables.php') ?? 'php';
 
-        if ($this->useAbsolutePathForMcp() || $forceAbsolutePath) {
+        if ($phpBinaryPath === 'php' && ($this->useAbsolutePathForMcp() || $forceAbsolutePath)) {
             return PHP_BINARY;
         }
 
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -177,17 +177,17 @@ class GuidelineAssist
 
     public function nodePackageManagerCommand(string $command): string
     {
-        $configuredNpm = config('boost.commands.npm');
+        $npmExecutable = config('boost.executables.npm');
 
-        if ($configuredNpm !== null) {
-            $nodePackageManagerCommand = $configuredNpm;
-        } elseif ($this->config->usesSail) {
-            $nodePackageManagerCommand = Sail::nodePackageManagerCommand($this->detectedNodePackageManager());
-        } else {
-            $nodePackageManagerCommand = $this->detectedNodePackageManager();
+        if ($npmExecutable !== null) {
+            return "{$npmExecutable} {$command}";
         }
 
-        return "{$nodePackageManagerCommand} {$command}";
+        if ($this->config->usesSail) {
+            return Sail::nodePackageManagerCommand($this->detectedNodePackageManager())." {$command}";
+        }
+
+        return "{$this->detectedNodePackageManager()} {$command}";
     }
 
     public function artisanCommand(string $command): string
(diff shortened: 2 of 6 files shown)


aivis:
I'm aware of https://github.com/laravel/boost/pull/419 , however, I think this more aligns with the existing approach (config variables are used in other places), later (if necessary) can be extended for NPM and other commands, and allows developers to have their settings in the boost config file.

aivis:
My reasoning for the `default` prefix is that these configuration variables do not guarantee specific outcome. 
Other high-level conditionals also affect the result, such as:

- `$this->useAbsolutePathForMcp() || $forceAbsolutePath`
- `$this->config->usesSail`

And they have higher precedence. 

These variables simply act as overrides or modifiers for the default bin paths.

aivis:
That said, I don't mind either way and I've made the necessary changes. It is probably a better config structure this way.

src/Install/CodeEnvironment/CodeEnvironment.php
```diff
@@ -40,7 +40,11 @@ public function useAbsolutePathForMcp(): bool
 
     public function getPhpPath(bool $forceAbsolutePath = false): string
     {
-        return ($this->useAbsolutePathForMcp() || $forceAbsolutePath) ? PHP_BINARY : 'php';
+        if ($this->useAbsolutePathForMcp() || $forceAbsolutePath) {
+            return PHP_BINARY;
+        }
+
+        return config('boost.commands.php', 'php');
```
pushpak1300: Precedence wise i think `config >> useAbsolutePathForMcp() (PHP_BINARY) >> php` should be the order ? 

If user is updating config that means they know what they are doing and wanted to override this path everywhere.

src/Install/GuidelineAssist.php
```diff
@@ -208,9 +208,13 @@ public function binCommand(string $command): string
...
-        return $this->config->usesSail
-            ? Sail::artisanCommand()
-            : 'php artisan';
+        if ($this->config->usesSail) {
+            return Sail::artisanCommand();
+        }
+
+        $phpBin = config('boost.commands.php', 'php');
+
+        return "{$phpBin} artisan";
```
pushpak1300: Precedence wise i think `config >> sail >> php` should be the order ? 

If user is updating config that means they know what they are doing and wanted to override this path everywhere.

src/Install/GuidelineAssist.php
fritz-c, line 226: I believe it would be useful to be able to customize bin command prefixes with config as well. For example, I have everything running in a docker container, so the command string I need is `docker compose exec app vendor/bin/{$command}`.

fritz-c:
Should some default config be added to https://github.com/laravel/boost/blob/1.x/config/boost.php to document this?

Also, could the existing `boost:install` command be tweaked to allow a user to modify these values at onboarding time?

src/Install/GuidelineAssist.php
aivis: Changed accordingly.

src/Install/CodeEnvironment/CodeEnvironment.php
aivis: Updated!

aivis:
> Should some default config be added to https://github.com/laravel/boost/blob/1.x/config/boost.php to document this?
> 
> Also, could the existing `boost:install` command be tweaked to allow a user to modify these values at onboarding time?

Adjusted the `boost.php` file. 

On the second point, I consider that out of scope. It could cause more confusion than help (?) and not everything is configurable through setup anyway.

src/Install/GuidelineAssist.php
pushpak1300: Thanks !

aivis:
@fritz-c @pushpak1300 Thanks for the reviews! I think the PR now covers all binaries + vendor bin prefix.

pushpak1300:
Thanks @aivis i will probably take a look today and try to merge.

pushpak1300:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

james2doyle:
I think it would be nice if it detected a `herd.yml` and ask if php and composer should be prefixed. But it sounds like that can be a new PR once this is in right?

smares:
Please consider #429 and that having the project set up with Sail does not mean every command needs to run via Sail, especially when using devcontainers where commands are ran "natively" without any prefix.

pushpak1300:
@aivis i will port this changes to main branch also. thanks for all the help.

PR #442, opened 2026-01-20
feat: add generic Docker support (without Sail)

marcemarin:
## Summary

This PR adds support for generic Docker Compose setups that don't use Laravel Sail. When `docker-compose.yml` or `compose.yaml` is detected without Sail being installed, the installer now prompts users to configure Boost MCP to run inside a Docker container.

**Problem:** Developers using custom Docker setups (not Laravel Sail) couldn't use Boost MCP effectively because the MCP server runs on the host machine but can't connect to services inside Docker containers (like databases with `DB_HOST=db`).

**Solution:** Similar to the existing Sail support, this PR:
- Detects Docker Compose files when Sail is not present
- Prompts users to configure Docker container execution
- Auto-detects common container names from `docker-compose.yml`
- Configures MCP to run via `docker exec -i <container> php artisan boost:mcp`

## Changes

- **New `Docker.php` class** (`src/Install/Docker.php`): Handles Docker container detection and command building, mirroring the Sail class structure
- **Config persistence** (`src/Support/Config.php`): Added `docker` and `docker_container` settings
- **Install command updates** (`src/Console/InstallCommand.php`):
  - Added Docker detection in `selectBoostFeatures()`
  - Added `shouldConfigureDocker()` confirmation prompt
  - Added `getDockerContainerName()` text input
  - Added `guessContainerName()` to auto-detect from docker-compose.yml
  - Updated `buildMcpCommand()` to use Docker when configured
  - Updated `installGuidelines()` to persist Docker config

## Test plan

- [x] Run existing test suite - all 385 tests pass
- [x] Tested locally with a custom Docker Compose setup (not using Sail)
- [x] Verified MCP can connect to containerized database after configuration
- [ ] Test on fresh project with Docker Compose but without Sail

Related to #429

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -15,6 +15,7 @@ use Laravel\Boost\Contracts\McpClient;
 use Laravel\Boost\Install\Cli\DisplayHelper;
 use Laravel\Boost\Install\CodeEnvironment\CodeEnvironment;
 use Laravel\Boost\Install\CodeEnvironmentsDetector;
+use Laravel\Boost\Install\Docker;
 use Laravel\Boost\Install\GuidelineComposer;
 use Laravel\Boost\Install\GuidelineConfig;
 use Laravel\Boost\Install\GuidelineWriter;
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks for your pull request!

For now, I’m going to hold off on merging this. To keep the package maintainable, we need to be careful about how much code and how many environment-specific cases we support. Since there are many possible Docker configurations, it’s not practical for us to detect and maintain all of them individually. A more generic approach, such as the one proposed in full request #437, seems more sustainable in the long run.

If you’d like, you could also consider releasing this as a separate package so the community can still benefit from your work.

PR #444, opened 2026-01-20
Add Skill Sync

pushpak1300:
### Summary

This PR improves how skills are managed by storing them as an array of skill names instead of a simple boolean. It also adds the ability to automatically clean up stale skills that are no longer defined, and fixes a bug where `false` config values were being incorrectly removed.

#### Skills stored as arrays
Previously, `boost.json` stored skills as `"skills": true/false`. Now it stores the actual skill names: `"skills": ["skill-one", "skill-two"]`. This allows Boost to know exactly which skills were managed by boost.

#### Automatic cleanup of stale skills
When running `boost:install --skills` or `boost:update`, skills that no longer exist in the source but are present in the target directory will be automatically removed. This keeps the skills directory clean when skills are renamed or removed.

The new `sync()` method in `SkillWriter` handles both writing new skills and removing stale ones in a single operation.

#### Config bug fix
Fixed an issue where setting `herd_mcp: false` (or any other `false` value) in `boost.json` would silently disappear after running `boost:update`. The config was incorrectly filtering out all "falsy" values when it should only remove `null` and empty arrays.

### DIFF
--- a/src/Concerns/RendersBladeGuidelines.php
+++ b/src/Concerns/RendersBladeGuidelines.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Laravel\Boost\Mcp\Prompts\Concerns;
+namespace Laravel\Boost\Concerns;
 
 use Illuminate\Support\Facades\Blade;
 use Laravel\Boost\Install\GuidelineAssist;
--- a/src/Install/GuidelineComposer.php
+++ b/src/Install/GuidelineComposer.php
@@ -6,8 +6,8 @@ namespace Laravel\Boost\Install;
 
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
+use Laravel\Boost\Concerns\RendersBladeGuidelines;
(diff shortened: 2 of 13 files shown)


src/Console/InstallCommand.php
```diff
@@ -434,6 +447,19 @@ protected function shouldUseSail(): bool
         return $this->selectedBoostFeatures->contains('sail');
     }
 
+    protected function isExplicitFlagMode(): bool
```
joetannenbaum, line 450: Isn't this just:

```php
return $this->option('guidelines') || $this->option('skills') || $this->option('mcp');
```

Or am I missing something?

pushpak1300: This is reverted by rector because of this rule `ReturnBinaryOrToEarlyReturnRector`

so all good there. do you want me to remove this rule ?

joetannenbaum: Nah all good, just a nit anyway.

PR #465, opened 2026-01-24
Feature/Support modular class discovery

soleinjast:
## Summary

This PR adds support for **modular Laravel project discovery** in Laravel Boost by consuming architecture signals exposed by `laravel/roster`.

By default, Boost only scans the `app/` directory. With this change, when Roster detects a **MODULAR** project structure, Boost will also discover classes located under a root-level `modules/` (or `Modules/`) directory.

---

## Motivation

Many Laravel applications use a modular architecture where domain code (such as Eloquent models, controllers, or enums) lives outside the `app/` directory.

Previously, Boost was unable to detect these classes, resulting in incomplete or misleading context during AI-assisted workflows.

This change delegates **architecture detection entirely to `laravel/roster`**, allowing Boost to remain simple and deterministic while supporting modular Laravel applications.

---

## Behavior

- Boost continues to scan the `app/` directory by default
- When Roster reports the `MODULAR` approach:
  - Boost additionally scans:
    - `modules/`
    - `Modules/` (case-insensitive)
- For all other project types, Boost falls back to the default behavior
- No filesystem heuristics or user configuration are required

---

## Related Work

- Project architecture detection is handled by `laravel/roster`
- Boost only consumes the detected approach and adjusts class discovery accordingly

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -25,26 +25,27 @@ return [
     | The following option may be used to enable or disable the browser logs
     | watcher feature within Laravel Boost. The log watcher will read any
     | errors within the browser's console to give Boost better context.
+    |
     */
 
     'browser_logs_watcher' => env('BOOST_BROWSER_LOGS_WATCHER', true),
 
     /*
     |--------------------------------------------------------------------------
-    | Boost Executables Config
+    | Boost Executables Paths
     |--------------------------------------------------------------------------
     |
-    | The following options allow you to configure custom paths for the PHP,
-    | Composer, and npm executables used by Boost. Leave empty to use defaults.
-    | When configured, these take precedence over automatic detection.
+    | These options allow you to specify custom paths for the executables that
+    | Boost uses. When configured, they take precedence over the automatic
+    | discovery mechanism. Leave empty to use defaults from your $PATH.
     |
     */
 
-    'executables' => [
-        'php' => env('BOOST_PHP_EXECUTABLE'),
-        'composer' => env('BOOST_COMPOSER_EXECUTABLE'),
-        'npm' => env('BOOST_NPM_EXECUTABLE'),
-        'vendor_bin' => env('BOOST_VENDOR_BIN_EXECUTABLE'),
+    'executable_paths' => [
+        'php' => env('BOOST_PHP_EXECUTABLE_PATH'),
+        'composer' => env('BOOST_COMPOSER_EXECUTABLE_PATH'),
+        'npm' => env('BOOST_NPM_EXECUTABLE_PATH'),
+        'vendor_bin' => env('BOOST_VENDOR_BIN_EXECUTABLE_PATH'),
     ],
 
 ];
--- a/src/Install/CodeEnvironment/CodeEnvironment.php
+++ b/src/Install/CodeEnvironment/CodeEnvironment.php
@@ -40,7 +40,7 @@ abstract class CodeEnvironment
 
     public function getPhpPath(bool $forceAbsolutePath = false): string
     {
-        $phpBinaryPath = config('boost.executables.php') ?? 'php';
+        $phpBinaryPath = config('boost.executable_paths.php') ?? 'php';
 
         if ($phpBinaryPath === 'php' && ($this->useAbsolutePathForMcp() || $forceAbsolutePath)) {
             return PHP_BINARY;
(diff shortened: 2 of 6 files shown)


pushpak1300:
@soleinjast I think instead of detecting the approach directly in boost we have laravel/roster which takes care of detecting approach and then we can use this here in boost app. 

Can you pleas send the PR first to properly detect the approach and then update rules and guidelines based on this here ?

soleinjast:
@pushpak1300 

Thanks for the clarification 🙌
Just to make sure I understand correctly:
should laravel/roster be responsible for detecting the project architecture / approach (e.g. default Laravel, modular, DDD), and then expose the detected source paths / signals for Boost to consume?

Also, for non-standard or custom architectures, is the expected behavior for Roster or better to say, Boost to fall back to the default strategy (e.g. app/) rather than trying to infer intent?

pushpak1300:
> should laravel/roster be responsible for detecting the project architecture / approach (e.g. default Laravel, modular, DDD), and then expose the detected source paths / signals for Boost to consume?

Yes

> Also, for non-standard or custom architectures, is the expected behavior for Roster or better to say, Boost to fall back to the default strategy (e.g. app/) rather than trying to infer intent?

I think in that case we can fallback to default. there are literally 100's of way laravel app can be customised and supporting each won't be possible.

soleinjast:
@pushpak1300 

just a heads-up — I’ve opened a PR in laravel/roster that adds first-class detection for **modular Laravel projects** via a root-level `modules/` directory and exposes this as a `MODULAR` approach.

This keeps architecture detection fully inside Roster, so Boost can consume the detected signals (approach / structure) without making filesystem assumptions or adding custom heuristics.

Roster PR: https://github.com/laravel/roster/pull/33

pushpak1300:
Thanks @soleinjast I will take a look at the the pr soon.

pushpak1300:
Hey @soleinjast I'm marking this MR as draft until it's ready. I have released the roster package. Thanks for all the help.

soleinjast:
> Hey @soleinjast I'm marking this MR as draft until it's ready. I have released the roster package. Thanks for all the help.

Thanks for the update!
I’ll update the Boost side to consume the released Roster signals and will mark the PR ready once that’s done.

soleinjast:
Hi @pushpak1300 👋

I’ve updated the PR to fully consume the released laravel/roster signals and removed the earlier configurable discovery paths approach.

Boost now relies on Roster for architecture detection (e.g. MODULAR) and falls back to the default app/ scan when no approach is detected, keeping all heuristics centralized in Roster as discussed.

Tests are updated, CI is green, and the PR should now be ready for review.

Thanks again for the guidance 🙌

webard:
Please also include the `app-modules` directory. It is used by https://github.com/InterNACHI/modular solution.

I created a Pull Request for Roster with `app-modules` support: https://github.com/laravel/roster/pull/36

pushpak1300:
hey @soleinjast can we target it on 2.x branch ?

Also check instead of doing this can we use reflection API ?

```php
 protected function classNameFromFile(string $path): ?string
    {
        $code = file_get_contents($path);

        if ($code === false) {
            return null;
        }

        $namespace = null;
        $class = null;

        $tokens = token_get_all($code);
        $count = count($tokens);

        for ($i = 0; $i < $count; $i++) {
            if (is_array($tokens[$i]) && $tokens[$i][0] === T_NAMESPACE) {
                $namespace = '';
                for ($j = $i + 1; $j < $count; $j++) {
                    if (is_array($tokens[$j]) && in_array($tokens[$j][0], [T_STRING, T_NS_SEPARATOR, T_NAME_QUALIFIED], true)) {
                        $namespace .= $tokens[$j][1];

                        continue;
                    }

                    if ($tokens[$j] === ';' || $tokens[$j] === '{') {
                        break;
                    }
                }
            }

            if (is_array($tokens[$i]) && in_array($tokens[$i][0], [T_CLASS, T_ENUM], true)) {
                for ($j = $i + 1; $j < $count; $j++) {
                    if (is_array($tokens[$j]) && $tokens[$j][0] === T_STRING) {
                        $class = $tokens[$j][1];
                        break 2;
                    }
                }
            }
        }

        if (! $class) {
            return null;
        }

        return $namespace ? "{$namespace}\\{$class}" : $class;
    }
```

soleinjast:
Hey @pushpak1300, about targeting correct branch — I have opened a new PR against it.
On ReflectionClass — it requires a class name, not a file path, so I think it can't replace the token parsing here. That step exists specifically to extract the class name before ReflectionClass can be used, which is already happening downstream in discover(). That said, I am open to suggestions if you have a cleaner approach in mind!

PR #472, opened 2026-01-26
Use skills directory instead of skill for OpenCode

thijsvdanker:
OpenCode expects its skill files to be in the `.opencode/skills` directory. See https://opencode.ai/docs/skills/#place-files


pushpak1300:
I think this had been changed [recently](https://github.com/anomalyco/opencode/commit/58f7da6e9f8fb08a9f42a0d2cc34a18b2475ddea#diff-84e0fe62561621d397c89a81a034e84f2f193e303f99f7914ba3a6f4affbacffL16) when i was testing it was still `/skill` it's still valid but make sense to rename it to skills.

PR #474, opened 2026-01-27
Sort user guidelines by filename for predictable ordering

patrickomeara:
I have been weighting my guidelines by filename `10-project.md`, `20-frontend.md` etc to give priority to the LLM.

Symfony's finder seems to let the OS decide which order the files are returned. My MacOS setup is returning the by last modified by default, which is changing the order whenever I update. 

This PR sorts the files by name before writing the agent files.


pushpak1300:
Thanks !

PR #478, opened 2026-01-27
Exclude Livewire skills and guidelines for indirect dependencies (#477)

### LINKED ISSUE #477: Skills added for indirect/transitive dependencies should be opt-in
### Boost Version

2.x (latest)

### Laravel Version

  12.48.1

### Agent/IDE Name & Version

Claude 2.1.20

### PHP Version

8.4.11

### System Info

macOS 26, Docker/Sail

### Description

 Laravel Boost adds skills for packages that are **indirect (transitive) dependencies**, not explicitly required in  
  `composer.json`. This can result in irrelevant skills being installed for packages the project doesn't actually use 
  directly.                                                                                                           
                                                                                                                      
  **Example**: A project using `laravel/pulse` gets the `livewire-development` skill installed, even though:          
  - Livewire is not in `composer.json`                                                                                
  - The project doesn't use Livewire directly                                                                         
  - Livewire is only present because Pulse depends on it                                                              
                                                                                                                      
  ```bash                                                                                                             
  $ composer why livewire/livewire                                                                                    
  laravel/pulse v1.5.0 requires livewire/livewire (^3.6.4|^4.0)
```                                                       
                                                                                                                      
  Current behavior: Skills are added for all detected packages, including transitive dependencies.                    
                                                                                                                      
  Expected behavior: Skills should only be added for direct dependencies by default, or there should be a             
  configuration option to control this.                                                                               
                                                                                                                      
  ### Suggested Solution                                                                                                  
                                                                                                                      
  Roster already has a $mustBeDirect array in DiscoverPackagePaths.php for packages that should only be included if   
  directly required:                                                                                                  
   ```                                                                                                                   
  protected array $mustBeDirect = [                                                                                   
      Packages::MCP,                                                                                                  
  ];                                                                                                                  
   ```                                                                                                                  
  Consider either:                                                                                                    
  1. Adding commonly transitive packages (like Livewire) to this list                                                 
  2. Making skills opt-in for indirect dependencies via boost.json configuration                                      
  3. Adding a directOnly option for skills in boost.json:
```                                                             
  {                                                                                                                   
      "skills": {                                                                                                     
          "directOnly": true                                                                                          
      }                                                                                                               
  }                                                                                                                   
   ```                                                                                                                   
  This would align with how Packages::MCP is already handled - only included when it's a direct requirement.                                                                                    

### Steps To Reproduce

                                                                                                                      
  1. Create a Laravel project with laravel/pulse (which requires Livewire):                                           
  composer require laravel/pulse                                                                                      
  2. Verify Livewire is NOT in your composer.json directly                                                            
  3. Install Laravel Boost:                                                                                           
  composer require laravel/boost --dev                                                                                
  php artisan boost:install                                                                                           
  4. Observe that livewire-development skill is installed despite not using Livewire directly

zcuric:
## Summary

  Fixes #477

  This PR adds `Packages::LIVEWIRE` to the `$mustBeDirect` array, ensuring that Livewire skills and guidelines are
  only included when Livewire is explicitly listed in the user's `composer.json`.

  ### Problem

  When a user installs a package like `laravel/pulse` that depends on Livewire as a transitive dependency, Boost was
  incorrectly including Livewire-specific skills and guidelines. This clutters the AI context with irrelevant
  information for users who aren't directly using Livewire.

  ### Solution

  Added Livewire to the existing `$mustBeDirect` mechanism (same approach used for `laravel/mcp`), so its
  skills/guidelines are only included when Livewire is a direct dependency.

  ### Changes

  - Added `Packages::LIVEWIRE` to `$mustBeDirect` array in `DiscoverPackagePaths.php`
  - Added tests for both `SkillComposer` and `GuidelineComposer` to verify indirect Livewire dependencies are excluded
  - Updated existing tests to explicitly mark Livewire as direct dependency where needed

  ### Benefit to End Users

  Users who install packages with Livewire as a transitive dependency (e.g., Laravel Pulse) will no longer receive
  Livewire-specific AI guidelines, resulting in cleaner, more relevant AI context.

  ### No Breaking Changes

  - Users who directly require `livewire/livewire` in their `composer.json` will continue to receive Livewire
  skills/guidelines as before
  - Only affects indirect/transitive dependencies

  ---

  ## Future Consideration

  The current `$mustBeDirect` approach requires manual code changes for each package. For a more scalable solution,
  consider one of these approaches:

  **Option 1: Default all packages to direct-only**
  ```php
  // Flip the logic - exclude ALL indirect packages by default
  return $package->indirect();

  Option 2: User-configurable via boost.json
  {
      "directOnly": ["livewire/livewire", "laravel/mcp"],
      "includeIndirect": ["some-package/i-want-anyway"]
  }

  Option 3: Opt-in for transitive dependencies
  {
      "includeTransitive": false,
      "allowTransitive": ["specific/package"]
  }

  This would give users control without requiring code changes for each new commonly-transitive package.
  ```

### DIFF
--- a/src/Install/Concerns/DiscoverPackagePaths.php
+++ b/src/Install/Concerns/DiscoverPackagePaths.php
@@ -19,6 +19,7 @@ trait DiscoverPackagePaths
      * */
     protected array $mustBeDirect = [
         Packages::MCP,
+        Packages::LIVEWIRE,
     ];
 
     /**
--- a/tests/Feature/Install/GuidelineComposerTest.php
+++ b/tests/Feature/Install/GuidelineComposerTest.php
@@ -355,6 +355,28 @@ test('includes laravel/mcp guidelines when directly required', function (): void
         ->not->toContain('Mcp::web');
 });
 
+test('excludes livewire guidelines when indirectly required', function (): void {
+    $packages = new PackageCollection([
+        new Package(Packages::LARAVEL, 'laravel/framework', '11.0.0'),
+        (new Package(Packages::LIVEWIRE, 'livewire/livewire', '3.0.0'))->setDirect(false),
+    ]);
+
+    $this->roster->shouldReceive('packages')->andReturn($packages);
+
+    expect($this->composer->compose())->not->toContain('=== livewire/core rules ===');
+});
+
+test('includes livewire guidelines when directly required', function (): void {
+    $packages = new PackageCollection([
+        new Package(Packages::LARAVEL, 'laravel/framework', '11.0.0'),
+        (new Package(Packages::LIVEWIRE, 'livewire/livewire', '3.0.0'))->setDirect(true),
+    ]);
+
+    $this->roster->shouldReceive('packages')->andReturn($packages);
+
+    expect($this->composer->compose())->toContain('=== livewire/core rules ===');
+});
+
 test('includes PHPUnit guidelines when Pest is not present', function (): void {
     $packages = new PackageCollection([
         new Package(Packages::LARAVEL, 'laravel/framework', '11.0.0'),
--- a/tests/Feature/Install/SkillComposerTest.php
+++ b/tests/Feature/Install/SkillComposerTest.php
@@ -22,7 +22,7 @@ beforeEach(function (): void {
 test('skills return a collection keyed by skill name', function (): void {
     $packages = new PackageCollection([
         new Package(Packages::LARAVEL, 'laravel/framework', '11.0.0'),
-        new Package(Packages::LIVEWIRE, 'livewire/livewire', '3.0.0'),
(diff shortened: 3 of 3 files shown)


pushpak1300:
> Option 1: Default all packages to direct-only 


I agree with this approach if there are not breaking change's. if you can send the pr i can review otherwise i will add this to my list. 

Thanks !

zcuric:
@pushpak1300 Thank you! I'll explore the Option 1 and submit the PR. 

Btw can you check this issues as well, becasue it's cause Interia skills not being added with boost. https://github.com/laravel/roster/pull/35

satoved:
I have the opposite problem with Intertia being included because of Laravel Spark.

PR #480, opened 2026-01-27
Add GitHub token configuration to avoid rate limits + improve request error handling

sebastiaanluca:
### Summary

Add support for GitHub Personal Access Tokens when fetching skills from remote repositories, and improve error handling to display meaningful API error messages instead of generic failures.

### Motivation

When using `boost:add-skill` to fetch skills from GitHub, users would hit rate limits on the unauthenticated GitHub API (60 requests/hour). The command would fail with a generic "No valid skills are found in the repository" message, giving no indication of what went wrong. This made debugging frustrating, especially for users unaware of GitHub's rate limiting.

### Changes

- Add `github_token` config option to authenticate GitHub API requests and avoid rate limiting
- Throw `GitHubSkillProviderException` with parsed error messages when API requests fail
- Display the actual GitHub API error message (e.g., "API rate limit exceeded...") instead of a generic failure message

### Remarks

Skipped a test because test run hanged in iTerm because the test expected input.

### DIFF
--- a/tests/Feature/Console/AddSkillCommandTest.php
+++ b/tests/Feature/Console/AddSkillCommandTest.php
@@ -164,7 +164,7 @@ it('skips existing skills without --force flag', function (): void {
     ])->assertSuccessful();
 
     $this->assertFileContains(['existing content'], '.ai/skills/skill-one/SKILL.md');
-})->skip(fn () => stream_isatty(STDIN), 'Skipped in interactive TTY environments');
+})->skip(fn (): bool => stream_isatty(STDIN), 'Skipped in interactive TTY environments');
 
 it('overwrites existing skills with --force flag', function (): void {
     File::ensureDirectoryExists(base_path('.ai/skills/skill-one'));


pushpak1300:
I think this is good.

most of my testing even with multiple skills i didn't face any issues. so for 90% of use case 60req/hr should be fine. 

I was thinking of the workflow. instead of asking token from api user in config can we somehow prompt user for token on failure ? 

Also if we face any rate limits can we somehow run `gh auth token` which will give us the token for the github user and then fetch the skills if we don't find this we can prompt it.

sebastiaanluca:
> most of my testing even with multiple skills i didn't face any issues. so for 90% of use case 60req/hr should be fine.

I added the vercel web skills for 3 projects which seems to have hundreds of skills and I hit it quite fast 😅 So it felt like it did multiple requests each time I added a skill, but not sure. Could've just been me adding too many repositories.

> I was thinking of the workflow. instead of asking token from api user in config can we somehow prompt user for token on failure ?

There's definitely room for optimization. Added the token and the simple error handling to avoid retrying without real feedback.

If we detect a GH rate limit and throw that exception, then explain and ask for the token, where should we store it? Temporarily for the current command? We can't write to .env to add it IMO. Also need to think of non-interactive mode so this makes the thing a bit more complex. We can handle that in this PR but we could also keep that for another iteration.

> Also if we face any rate limits can we somehow run `gh auth token` which will give us the token for the github user and then fetch the skills if we don't find this we can prompt it.

Didn't know that was a thing! But needs a check if the `gh` command exists.

pushpak1300:
Another approach is to use trees api https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28 which we can call 60 times - tried locally for all vercel skills for 20 times and hasn't hit the rate limit yet. `raw.githubusercontent.com/*` has more than 5000 req/hr rate limit.

All other sort of optimisation can be done on top of this i am avoiding to take gh token from user unless very necessary. I will update the changes on your branch. Thanks.

pushpak1300:
Thanks for all the help created #520 with the tree api.

PR #483, opened 2026-01-27
add `boost:rm-skill` command to remove installed skills

dominosaurs:
hi maintainers,

personally, i needed this command because as the number of installed skills grows, manually managing files inside `.ai/skills` becomes annoying and easy to mess up. having a first-class, explicit removal command makes day-to-day skill management much smoother and safer for me, and i thought it might be useful for others as well.

if this turns out to be a bad idea, or doesn’t align with the project’s direction, that’s totally fine and i’m okay with this being rejected.

it supports both interactive multi-selection and direct arguments for specifying skills. Includes safety confirmations that can be bypassed using a force flag, and automatically triggers a system update after removal to maintain consistency.

#### usage

```bash
# remove specific skills
php artisan boost:rm-skill skill-one skill-two

# remove with interactive selection
php artisan boost:rm-skill

# force remove without confirmation
php artisan boost:rm-skill skill-one --force
```

feedback is very welcome, especially if there’s a better mental model or workflow you’d prefer for skill removal.

### DIFF
--- a/src/BoostServiceProvider.php
+++ b/src/BoostServiceProvider.php
@@ -107,6 +107,7 @@ class BoostServiceProvider extends ServiceProvider
                 Console\UpdateCommand::class,
                 Console\ExecuteToolCommand::class,
                 Console\AddSkillCommand::class,
+                Console\RemoveSkillCommand::class,
             ]);
         }
     }
--- a/src/Console/RemoveSkillCommand.php
+++ b/src/Console/RemoveSkillCommand.php
@@ -1,0 +1,185 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Boost\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\File;
+use Laravel\Boost\Concerns\DisplayHelper;
+use Laravel\Prompts\Terminal;
+
+use function Laravel\Prompts\confirm;
+use function Laravel\Prompts\multiselect;
+use function Laravel\Prompts\spin;
+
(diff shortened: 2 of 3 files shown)


zainphp:
"it works on my machine" moment 😞

```bash
   PASS  Tests\Feature\Console\RemoveSkillCommandTest
  ✓ it removes a single skill with confirmation                                                                                                    0.47s  
  ✓ it removes multiple skills with confirmation                                                                                                   0.16s  
  ✓ it cancels removal when confirmation is denied                                                                                                 0.07s  
  ✓ it removes skills without confirmation using --force                                                                                           0.12s  
  ✓ it fails to remove a non-existent skill                                                                                                        0.07s  
  ✓ it prompts for skill selection if no argument is provided                                                                                      0.18s  
  ✓ it displays error when no skills are installed when prompting                                                                                  0.08s  

  Tests:    7 passed (29 assertions)
  Duration: 1.47s
  ```

pushpak1300:
Thanks for sending PR. Don't you think instead of doing this - you can easily delete multiple skills and run `boost:update` ?

zainphp:
yeah, that’s fair. but for me personally, using a command still feels easier and safer than deleting things manually and then remembering to run `boost:update` afterward.

but of course, this is just how it feels in my head, not an objective requirement. totally fine if this doesn’t make sense for the project or if you decide not to accept it ❤️

pushpak1300:
Okay if this comes up multiple time we can take a look at it and see if this should bepart of boost. currently i feel it's just not gonna be useful. 

Thanks for the PR ❤️

PR #485, opened 2026-01-28
add Laravel Valet support

feenx:
Laravel Valet support, determined by question during install. Extends Laravel Sail dynamic artisan/composer command generation.

I saw some PRs about adding composer/php binary paths, but it seemed cleaner to me just add first party support for Laravel Valet similar to Laravel Sail's implementation. Albeit with the confirmation question, as each project has nothing to determine if Laravel Valet is being used.

Changes:
- Introduce Valet-specific guideline configurations.
- Update commands to recognize Valet for composer and artisan.
- Extend tests to verify Valet functionality.
- Add usesValet property across relevant classes.
- Update Config to manage Valet settings.

### DIFF
--- a/.ai/fluxui-free/core.blade.php
+++ b/.ai/fluxui-free/core.blade.php
@@ -1,4 +1,4 @@
 # Flux UI Free
 - Flux UI is the official Livewire component library. This project uses the free edition, which includes all free components and variants but not Pro components.
-- Use `<flux:*>` components when available; they are the recommended way to build Livewire interfaces.
+- Use @verbatim`<flux:*>`@endverbatim components when available; they are the recommended way to build Livewire interfaces.
 - IMPORTANT: Activate `fluxui-development` when working with Flux UI components.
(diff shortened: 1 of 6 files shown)


src/Console/InstallCommand.php
```diff
@@ -217,6 +224,15 @@ protected function shouldConfigureSail(): bool
...
 
+    protected function shouldUseValet(): bool
+    {
+        return confirm(
+            label: 'Is this project running under Laravel Valet?',
+            default: $this->config->getValet(),
+            hint: 'This will add Valet-specific command guidelines (e.g. `valet php artisan`).',
+        );
+    }
+
```
pushpak1300, line 235: Ideally, if the user doesn’t have Valet installed, we shouldn’t ask this question at all.

pushpak1300:
Valet support can also be added via #491, so instead of asking more questions during installation (which hurts the UX), users who explicitly need Valet support can simply enable it by adding this environment variable.
```
BOOST_PHP_EXECUTABLE="valet php"
BOOST_COMPOSER_EXECUTABLE="valet composer"
```

feenx:
Considering we ask any questions at all, it seems counterintuitive to require Valet users to find out it doesn't work, and go find env options such as your PR has. Verus the default of No during setup.

Perhaps you know market share information, regarding Valet usage, as I do not.

PR #494, opened 2026-01-28
Add Flux Icons documentation to SKILL.md files

### LINKED ISSUE #177: Flux Docs missing info about icons
It seems that we're lacking context about icon options in Flux. The AI keeps making up icon names that do not actually exist.  For example:

```
 <flux:button icon="download" wire:click="export">{{ __('Export to Excel') }}</flux:button>
````

There is no "download" icon. The closest would be "arrow-down-tray."

Now, the _Boost docs do not actually include a list of available icon names_. It just points to https://heroicons.com/ for a list of available values. Perhaps we need to list all available Heroicons and pass that list to Boost.

### PR DESCRIPTION
**Pull request by pushpak1300**
Added Icons section to both `fluxui-free` and `fluxui-pro` SKILL.md files documenting Heroicons as the default icon set and Lucide as an alternative addresses issue where AI assistants invent non-existent icon names (e.g., download instead of arrow-down-tray)

Fixes #177

### DIFF
--- a/.ai/fluxui-free/skill/fluxui-development/SKILL.md
+++ b/.ai/fluxui-free/skill/fluxui-development/SKILL.md
@@ -32,6 +32,10 @@ Use Flux UI components when available. Fall back to standard Blade components wh
 <flux:button variant="primary">Click me</flux:button>
 </code-snippet>
 
+## Available Components (Free Edition)
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/Pest.php [hunk 1]
```diff
@@ -21,43 +21,43 @@
 
 expect()->extend('isToolResult', fn () => $this->toBeInstanceOf(Response::class));
 
-expect()->extend('toolTextContains', function (mixed ...$needles) {
+expect()->extend('toolTextContains', function (mixed ...$needles): object {
```
pushpak1300, line 24: The changes been done by rector.

Issue #466, opened 2026-01-25
database-schema call: too many tokens (gets ALL databases)


PovilasKorop:
### Boost Version

1.8.10

### Laravel Version

12

### Agent/IDE Name & Version

Claude Code 2.1.19

### PHP Version

8.4

### System Info

MacOS

### Description

When running prompts for longer tasks with Laravel Boost, I noticed these warnings happening quite often in the **beginning** of the process, in the research phase of Claude Code work.

<img width="1727" height="463" alt="Image" src="https://github.com/user-attachments/assets/8d7e6c33-c422-43f3-80d3-6cc41eea04ca" />

<img width="1162" height="447" alt="Image" src="https://github.com/user-attachments/assets/0a524892-dc13-4e3b-9540-b5b61dd91b04" />

So some MCP call immediately took like ~10% of context window.

I didn't dig what tool/MCP caused so many tokens. Until now.

It's Laravel Boost `database-schema` call.

I opened the process details with `ctrl + o` and noticed the return of `database-schema` was a huge JSON with **all tables of my MySQL server, including other local databases**.

---

As a proof, I also added this at the end of the prompt to Claude Code:

Work on phase 3.1 from @docs/tasks.md - when done, list the MCP tools calls with the amount of tokens they used. I want to understand which actions take the most tokens and how to optimize the context management.

Here's the result:

```
  MCP Tool Calls Token Analysis
  Tool: mcp__laravel-boost__database-schema
  Calls: 1
  Estimated Token Usage: ~8,000-10,000
  Notes: Largest - returned full schema for 150+ tables
  ────────────────────────────────────────
  Tool: mcp__laravel-boost__application-info
  Calls: 1
  Estimated Token Usage: ~400
  Notes: Moderate - packages, versions, models
  ────────────────────────────────────────
  Tool: mcp__laravel-boost__database-query
  Calls: 2
  Estimated Token Usage: ~800 each
  Notes: Moderate - returned seeded tag data
  Optimization Recommendations:

  1. database-schema - This was the heaviest call. Use the filter parameter to limit to specific tables (e.g.,
  filter: "recipe") when you only need certain tables
  2. application-info - Already efficient. The models array was empty, but could grow large in bigger apps
  3. database-query - Use LIMIT clauses when you just need to verify data exists rather than see all records
  4. General - For new features involving specific tables, prefer:
    - Reading existing model files directly (minimal tokens)
    - Filtering database-schema to relevant tables
[...]

PR #498, opened 2026-01-29
Fix database-schema tool to only return tables from configured database

### LINKED ISSUE #202: API error every time Claude accesses the database schema
### Laravel Package Version

1.0.18

### Laravel Version

11.x

### PHP Version

8.4

### System Info

PopOS 22.044

### Description

Hello!

We have hundreds of tables and view, and when I ask claude to look at a particular schema I usually get an error along the lines of:

```
AI_APICallError: input length and `max_tokens` exceed context limit: 188908 + 32000 > 200000, decrease input length or `max_tokens` and try again
```

or
 
```
Error: MCP tool "database-schema" response (85478 tokens) exceeds maximum allowed tokens (25000). Please use pagination, filtering, or limit parameters to reduce the response size.
```

### Steps To Reproduce

ask boost to look at the database schema

pushpak1300:
Fixes #202 and #466 by ensuring the `database-schema` tool only returns tables from the configured Laravel database and reducing default response size.

**Issue #466**: The `database-schema` tool was returning tables from ALL databases on the MySQL server, not just the configured Laravel database. Users with 10 tables in their project were seeing 150+ tables from other local databases.

**Issue #202**: The response was too large (85k+ tokens), exceeding MCP context limits and consuming ~10% of the context window in a single call.

### Root Cause

The tool was using Laravel's `Schema::getTables()` which relies on MySQL's `SHOW TABLES` command. This returns all tables the MySQL user has access to, which in shared hosting or local development environments can include tables from multiple databases.

Meanwhile, other methods in the MySQL driver (views, procedures, functions) were correctly filtering with `WHERE TABLE_SCHEMA = DATABASE()`, but the main table listing was not.

### 1. Database-Scoped Table Queries

Added a `getTables()` method to each database driver with explicit database filtering:

| Driver | Filter |
|--------|--------|
| MySQL | `WHERE TABLE_SCHEMA = DATABASE()` |
| PostgreSQL | `WHERE schemaname = current_schema()` |
| SQLite | Queries `sqlite_master` (inherently scoped) |

### 2. Opt-in for Views & Routines

Changed the response structure to exclude views and routines by default:

- Added `include_views` parameter (default: `false`)
- Added `include_routines` parameter (default: `false`)

This significantly reduces response size for the common case where only table schemas are needed.

### Response Structure

**Default response (minimal):**
```json
{
  "engine": "mysql",
  "tables": { ... }
}
```

**With `include_views: true, include_routines: true`:**
```json
{
  "engine": "mysql",
  "tables": { ... },
  "views": [ ... ],
  "routines": {
    "stored_procedures": [ ... ],
    "functions": [ ... ],
    "sequences": [ ... ]
  }
}
```

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema/DatabaseSchemaDriver.php
+++ b/src/Mcp/Tools/DatabaseSchema/DatabaseSchemaDriver.php
@@ -6,7 +6,10 @@ namespace Laravel\Boost\Mcp\Tools\DatabaseSchema;
 
 abstract class DatabaseSchemaDriver
 {
-    public function __construct(protected ?string $connection = null) {}
+    public function __construct(protected ?string $connection = null)
+    {
+        //
+    }
 
     protected function hasTable(?string $table): bool
     {


calebdw:
@pushpak1300 this is good, but we have like 300-400 tables in one database, i think there probably needs to be some filtering done before sending the entire schema over the wire (or maybe use TOON instead of json)

PR #499, opened 2026-01-29
Add symlink install mode for skills

hosmelq:
## Summary

- Add optional symlink install mode for skills with canonical `.agents/skills` storage and copy fallback.

## References

- Uses the same canonical `.agents/skills` location as the Vercel [skills](skills.sh) CLI.

| Before | After |
|--------|--------|
| CleanShot 2026-01-29 at 11 12 12@2x | CleanShot 2026-01-29 at 11 12 17@2x |

### DIFF
--- a/src/Install/SkillWriter.php
+++ b/src/Install/SkillWriter.php
@@ -150,12 +150,13 @@ class SkillWriter
                 return true;
             }
 
-            // On Windows, directory symlinks can require rmdir instead of unlink.
-            if (is_dir($path) && @rmdir($path)) {
+            // On Windows, directory symlinks can require rmdir instead of unlink,
+            // even when the symlink target no longer exists (dangling symlinks).
+            if (@rmdir($path)) {
                 return true;
             }
 
-            return ! file_exists($path);
+            return ! file_exists($path) && ! is_link($path);
         }
 
         if (is_file($path)) {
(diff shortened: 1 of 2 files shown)


pushpak1300:
So for now instead of adding this as optional mode we can make it default one but at the same time it should be looking at `.ai` folder and not the `.agents`.

hosmelq:
@pushpak1300, these changes have been applied.

pushpak1300 (approved):

Plytas:
@pushpak1300 is there a way to restore previous behaviour and have the skills be actual files, not symlinks?

pushpak1300:
Nope at the moment. Is there any issues you facing with the symlink mode @Plytas  ?

Plytas:
@pushpak1300 thanks for taking your time to reply.

I'm having issues with Devin picking up skills. After some research I was led to believe that symlinks might be a problem. However, I now believe it's a Devin issue. I'm unable to index a repo, which should discover skills. Even an unrelated repo without symlinks is failing to index.

Kovah:
I updated Boost to the latest version and have Codex, Copilot and Junie enabled.
After removing `.github/skills` and `.junie/skills`, running `boost:update` keeps creating files all over the place, instead of symlinking `.github/skills` and `.junie/skills` to `.agent/skills`. How is symlinking supposed to work at the moment?

Plytas:
@pushpak1300 looks like Devin fixed indexing issues and not it is able to discover skills. However it seems that it doesn't detect symlinked skils, only the actual files.

Would it be possible to implement an option that would copy actual files and not make symlinks?

Issue #503, opened 2026-01-30
Gemini CLI error when parsing `@` in generated guidelines markdown


dominosaurs:
### Boost Version

2.0.4

### Laravel Version

12.49.0

### Agent/IDE Name & Version

Gemini CLI 0.26.0

### PHP Version

8.4.11

### System Info

Windows 11

### Description

when using the guidelines ([context file/GEMINI.md](https://geminicli.com/docs/cli/gemini-md/)) generated by `laravel/boost` with Gemini CLI. when Gemini starts, it shows errors because it interprets `@` in markdown as [a local directory reference](https://geminicli.com/docs/cli/commands/#at-commands).

this was also surprising to me because the 'at command' is also applies inside markdown content.

my current workaround is manually escaping `@` with a backslash (`\@`) and errors gone. i wish consider automatically escaping `@` in the generated guidelines file for Gemini only on `Foundational Context` section.

generated `GEMINI.md`:

```md
<!-- .... -->
## Foundational Context
<!-- .... -->
- inertiajs/inertia-laravel (INERTIA) - v2
- laravel/wayfinder (WAYFINDER) - v0
- @inertiajs/react (INERTIA) - v2
- @laravel/vite-plugin-wayfinder (WAYFINDER) - v0
<!-- .... -->
```

error example:

```
✖  [ERROR] [ImportProcessor] Failed to import inertiajs/react: ENOENT: no such file or directory, access 'D:\projects\my-app\inertiajs\react'
✖  [ERROR] [ImportProcessor] Failed to import laravel/vite-plugin-wayfinder: ENOENT: no such file or directory, access 'D:\projects\my-app\laravel\vite-plugin-wayfinder'
```

### Steps To Reproduce

- install dependency packages that use `@` in the name.
- run `boost:update`
- start Gemini CLI

PR #523, opened 2026-02-04
fix(gemini): escape @ symbols in foundational context

soleinjast:
### What changed
Gemini CLI interprets `@scope/package` strings inside Markdown as import directives, attempting to resolve them as local filesystem paths. This causes `ENOENT` errors when reading the generated `GEMINI.md` guidelines file.

This PR introduces a **Gemini-specific post-processing step** that escapes `@` symbols **only in the “Foundational Context” section** of the generated guidelines.

### Why this is needed
- `@scope/package` is valid and common npm syntax
- Gemini applies import parsing even inside Markdown content
- Escaping `@` prevents unintended import resolution
- The behavior is limited to Gemini and does not affect other agents

### How it works
- Added an optional `PostProcessGuidelines` contract
- Implemented the contract for the Gemini agent only
- Applied post-processing in `GuidelineWriter` when supported
- Included tests to ensure scoped package escaping and idempotency

### Scope & safety
- **Agent-specific:** Gemini only
- **Section-scoped:** Foundational Context only
- **Non-breaking:** No changes for other agents or existing workflows

### Related
Fixes https://github.com/laravel/boost/issues/503

### DIFF
--- a/src/Contracts/SupportsGuidelines.php
+++ b/src/Contracts/SupportsGuidelines.php
@@ -20,7 +20,7 @@ interface SupportsGuidelines
     public function frontmatter(): bool;
 
     /**
-     * Post-process the generated guidelines' Markdown.
+     * Transform the generated guidelines' Markdown.
      */
-    public function postProcessGuidelines(string $markdown): string;
+    public function transformGuidelines(string $markdown): string;
 }
--- a/src/Install/Agents/Agent.php
+++ b/src/Install/Agents/Agent.php
@@ -239,7 +239,7 @@ abstract class Agent
     /**
      * Post-process the generated guidelines' Markdown.
      */
-    public function postProcessGuidelines(string $markdown): string
+    public function transformGuidelines(string $markdown): string
     {
         return $markdown;
     }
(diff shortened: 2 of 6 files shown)


src/Contracts/PostProcessGuidelines.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Laravel\Boost\Contracts;
+
+interface PostProcessGuidelines
+{
+    /**
+     * Post-process the generated guidelines markdown.
+     */
+    public function postProcessGuidelines(string $markdown): string;
+}
```
pushpak1300: don't you think we don;t need this interface ? we could define the method in `Agent` with null or callback which can handle this post processing logic automatically ?

soleinjast: @pushpak1300  Initially, I introduced a separate interface to keep the Agent contract minimal and make guideline post-processing explicitly opt-in, since it’s a very agent-specific concern and not something most agents would need.

After your comment, I agree this can be simplified. I’ve refactored the implementation by moving postProcessGuidelines() onto the base Agent as a no-op default and overriding it only in Gemini.

Let me know if any change needed!

It's ready for review!

Issue #316, opened 2025-10-24
Command database-query doesn't respect mysql prefix correctly


braseidon:
### Laravel Package Version

1.4

### Laravel Version

12.35.1

### PHP Version

8.4.13

### System Info

Windows 11, Herd, Claude Code

### Description

Hi. When Claude Code queries the database using database-query, it gives a MySQL error. This is because my app uses a MySQL prefix for its tables. The query is using a period instead of an underscore for the prefix.

My `config/database.php`:

```
return [
        'mysql' => [
            'prefix' => 'arpg_',
        ],
];
```

And the database query:

```
laravel-boost - database-query (MCP)(query: "SELECT * FROM poe1_leagues ORDER BY startAt DESC LIMIT 5")
  ⎿  Error: Query failed: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'arpg.poe1_leagues' doesn't exist (Connection: mysql, SQL: SELECT * FROM poe1_leagues ORDER BY startAt DESC LIMIT 5)
```

### Steps To Reproduce

Use a prefix for MySQL tables using an underscore `prefix_` and then have Claude query the database.

sakshamgorey:
Hey @pushpak1300, I noticed this issue is assigned to you. If you’re already working on it, I’m happy to help with a repro or add test coverage. If you haven’t started yet (or don’t mind), I’d love to take a shot at a fix and open a PR.

pushpak1300:
@sakshamgorey i haven't yet happy to assign it to you.

braseidon:
I've had this issue happen numerous times since I made this issue. If you'd like, I can search my past conversations to find examples

sakshamgorey:
Hey @braseidon,I’ve been facing this same challenge in my own projects.

Currently, I’ve been circumventing this with two workarounds:

1. Instructional: In my agent.md file, I explicitly state the table prefix and instruct the AI to apply it manually to the database query tool.
2. Manual Tool Update: I previously modified the vendor package to update the tool's parameter description and make the connection field required to force the prefixing(overkill).

I’m working on a more robust solution that ensures the tool automatically respects the prefix defined in config/database.php. While I’m taking the lead on this fix, I’m definitely open to suggestions or collaboration if anyone else has a specific implementation in mind.

pushpak1300:
Released this in https://github.com/laravel/boost/releases/tag/v2.1.2

PR #529, opened 2026-02-05
fix: apply table prefix to raw SQL queries in DatabaseQuery tool

soleinjast:
## Problem
The `DatabaseQuery` MCP tool doesn't respect Laravel's table prefix configuration when executing raw SQL queries. When apps use a prefix like `'prefix' => 'arpg_'`, queries fail with "table not found" errors.

## Solution
This PR adds automatic prefix application:
- Retrieves prefix using `$connection->getTablePrefix()`
- Adds `addPrefixToQuery()` method with regex to match table names after FROM, JOIN, INTO, UPDATE, TABLE keywords

**Example:**
```sql
-- Input (with prefix 'arpg_')
SELECT * FROM poe1_leagues
-- Output
SELECT * FROM arpg_poe1_leagues
```

## Benefits
- Fixes queries for apps using table prefixes
- Backward compatible (no impact when prefix is empty)
- No breaking changes

## Testing
- Updated tests to mock `getTablePrefix()`
- All existing tests pass

### DIFF
--- a/.ai/boost/core.blade.php
+++ b/.ai/boost/core.blade.php
@@ -10,6 +10,7 @@
 ## Tinker / Debugging
 - You should use the `tinker` tool when you need to execute PHP to debug code or query Eloquent models directly.
 - Use the `database-query` tool when you only need to read from the database.
+- Use the `database-schema` tool to inspect table structure before writing migrations or models.
 
 @if (config('boost.browser_logs', true) !== false || config('boost.browser_logs_watcher', true) !== false)
 ## Reading Browser Logs With the `browser-logs` Tool
--- a/.ai/fluxui-free/skill/fluxui-development/SKILL.md
+++ b/.ai/fluxui-free/skill/fluxui-development/SKILL.md
@@ -1,10 +1,9 @@
 ---
 name: fluxui-development
-description: >-
-  Develops UIs with Flux UI Free components. Activates when creating buttons, forms, modals,
-  inputs, dropdowns, checkboxes, or UI components; replacing HTML form elements with Flux;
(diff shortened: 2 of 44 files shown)


sakshamgorey:
Hey @soleinjast , thanks for the PR ,this is a solution for #316 and I ran into a blocker with this exact approach regarding **CTE queries**.

The issue is that the logic blindly prefixes table names, so it incorrectly prefixes CTE aliases too.

**Example:**
```sql
WITH recent_users AS (
    SELECT * FROM users
)
SELECT * FROM recent_users
```

If a prefix is active (e.g., `app_`), this results in `SELECT * FROM app_recent_users`, which throws a "Table not found" error. We might need a more robust way to exclude `WITH` aliases.

soleinjast:
> Hey @soleinjast , thanks for the PR ,this is a solution for #316 and I ran into a blocker with this exact approach regarding **CTE queries**.
> 
> The issue is that the logic blindly prefixes table names, so it incorrectly prefixes CTE aliases too.
> 
> **Example:**
> 
> ```sql
> WITH recent_users AS (
>     SELECT * FROM users
> )
> SELECT * FROM recent_users
> ```
> 
> If a prefix is active (e.g., `app_`), this results in `SELECT * FROM app_recent_users`, which throws a "Table not found" error. We might need a more robust way to exclude `WITH` aliases.

Hey @sakshamgorey!
Exactly - the issue was that the logic didn't differentiate between real tables and CTE aliases. The updated implementation now identifies CTE names first and skips them during the prefixing process, which resolves the "Table not found" errors you encountered.

pushpak1300:
Thanks

PR #533, opened 2026-02-06
Graceful cache fallback when cache driver is unreachable

pushpak1300:
When the configured cache driver is unreachable (e.g. database driver with PostgreSQL down, Redis connection refused), Boost throws unhandled exceptions like QueryException. This crashes boost:install, DatabaseSchema, and LastError MCP tools — none of which actually require cache to function.                                                                       
                                                                                                                                                                                           
### Solution                                                                                                                                                                            
Wrapped all cache operations in Laravel's rescue() helper with report: false. Each call site has a natural fallback:
  - BoostServiceProvider — cache miss triggers fresh Roster::scan(), failed write is skipped
  - DatabaseSchema — falls back to direct getDatabaseStructure() call
  - LastError — falls back to reading the log file on disk

  Added tests to verify LastError and DatabaseSchema degrade gracefully when cache fails.

### DIFF
--- a/.ai/inertia-laravel/core.blade.php
+++ b/.ai/inertia-laravel/core.blade.php
@@ -1,7 +1,7 @@
 # Inertia
 
 - Inertia creates fully client-side rendered SPAs without modern SPA complexity, leveraging existing server-side patterns.
-- Components live in `resources/js/Pages` (unless specified in `vite.config.js`). Use `Inertia::render()` for server-side routing instead of Blade views.
+- Components live in `{{ $assist->inertia()->pagesDirectory() }}` (unless specified in `vite.config.js`). Use `Inertia::render()` for server-side routing instead of Blade views.
 - ALWAYS use `search-docs` tool for version-specific Inertia documentation and updated code examples.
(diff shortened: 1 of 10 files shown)


src/BoostServiceProvider.php
```diff
@@ -47,17 +47,18 @@ public function register(): void
             $cacheKey = 'boost.roster.scan';
             $lastModified = max(array_map(fn (string $path): int|false => file_exists($path) ? filemtime($path) : 0, $lockFiles));
 
-            $cached = cache()->get($cacheKey);
+            $cached = rescue(fn () => cache()->get($cacheKey), report: false);
```
siarheipashkevich, line 50: @pushpak1300 why are you mixing `Cache::` facade and the `cache()->` helper?

pushpak1300: @siarheipashkevich ?

siarheipashkevich: In the files above you are using facade cache, but here helper method. Are there any reasons why you use different ways?

pushpak1300: No reasons. will fix it.

PR #537, opened 2026-02-06
feat: allow extending guidelines/skills with custom content

### LINKED ISSUE #426: Extend guidelines instead of replacing them
Overriding guidelines works fine, but sometimes I'd like to append some instructions. 
In my case for example I'd like to add an instruction after the  pest/core rules, leaving the original guideline untouched.

My instruction would be this: "Only write tests if the user explicitly requests you to do so."
This way I can use the pest rules but only if needed.

I tried adding this by creating a custom guideline in .ai/guidelines/testing. After boost:update this custom guideline is put at the top of the generated guidelines.md, but it's ignored completely. My coding agent (junie) writes tests anyway. Maybe because the pest rules are read after my instruction and considered more important?

### LINKED ISSUE #536: Feature Request: allow users to EXTEND guidelines and skills, not just overwrite
Hello!

Like the title says, it would be **awesome** to allow **extending** a guideline or skill instead of just completely overwriting it. This is **critical**---I need to be able to add custom conventions/instructions/etc. and I don't want to have to completely overwrite and basically copy/paste from the existing guideline/skill because I still want that stuff there.

### Recommendation

When Boost loads a guideline or skill, it also checks for extension files with `.append.md` or `.prepend.md` suffixes and merges them automatically.

#### For guidelines:
.ai/guidelines/
├── laravel/11/core.append.md      # Appended to built-in Laravel 11 guidelines
├── pest/core.prepend.md           # Prepended to built-in Pest guidelines
└── my-custom-rules.md             # Standalone guideline (existing behavior)

#### For skills:
.ai/skills/
├── pest-testing/SKILL.append.md   # Appended to built-in pest-testing skill
└── my-custom-skill/SKILL.md       # Standalone skill (existing behavior)

#### Behavior
- `*.append.md`: content is appended after the built-in guideline/skill
- `*.prepend.md`: content is prepended before the built-in guideline/skill
- No suffix (existing behavior) - Fully replaces the built-in

#### Example

```md
<!-- .ai/skills/pest-testing/SKILL.append.md -->
### Project Testing Conventions
#### Naming
- Prefix integration tests with `Integration`
#### Database
- Always use `RefreshDatabase` trait, never `DatabaseTransactions`
- Seed test-specific data in the test, not via seeders
#### Assertions
- Prefer `assertDatabaseHas` over manual queries
- Use custom assertions in `tests/Assertions` for domain-specific checks
```

Thanks!

calebdw:
Hello!

Closes #536, closes #426 

Users can now extend built-in guidelines and skills without replacing them entirely by using `*.append.md` and `*.prepend.md` files:

#### Guidelines
- `.ai/guidelines/{path}/core.append.md`: appends to built-in guideline
- `.ai/guidelines/{path}/core.prepend.md`: prepends to built-in guideline

#### Skills
- `.ai/skills/{skill-name}/SKILL.append.md`: appends to built-in skill
- `.ai/skills/{skill-name}/SKILL.prepend.md`: inserts after frontmatter

This allows users to add project-specific conventions, examples, and context while still receiving updates to Boost's maintained guidelines and skills.

> [!TIP]
> Both `.blade.php` and `.md` extensions are supported for extension files.

CC: @cswgr

Thanks!

### DIFF
--- a/src/Install/GuidelineComposer.php
+++ b/src/Install/GuidelineComposer.php
@@ -255,6 +255,7 @@ class GuidelineComposer
         }
 
         $rendered = $this->renderBladeFile($path);
+        $rendered = $this->applyGuidelineExtensions($path, $rendered);
 
         $description = Str::of($rendered)
             ->after('# ')
@@ -273,6 +274,56 @@ class GuidelineComposer
             'third_party' => $thirdParty,
             'tokens' => round(str_word_count($rendered) * 1.3),
         ];
+    }
+
+    protected function applyGuidelineExtensions(string $basePath, string $content): string
+    {
+        if (str_contains($basePath, $this->customGuidelinePath())) {
+            return $content;
+        }
+
+        $relativePath = $this->getRelativeGuidelinePath($basePath);
+
+        if ($relativePath === null) {
+            return $content;
+        }
+
+        $prependContent = $this->getExtensionContent($relativePath, 'prepend');
+        $appendContent = $this->getExtensionContent($relativePath, 'append');
+
+        if ($prependContent !== '') {
+            $content = trim($prependContent)."\n\n".$content;
+        }
+
+        if ($appendContent !== '') {
+            $content = $content."\n\n".trim($appendContent);
+        }
+
+        return $content;
+    }
+
+    protected function getRelativeGuidelinePath(string $absolutePath): ?string
+    {
+        $basePath = realpath(__DIR__.'/../../');
+
+        return Str::of($absolutePath)
+            ->replace([$basePath, '.ai'.DIRECTORY_SEPARATOR, '.ai/'], '')
+            ->ltrim('/\\')
+            ->replace(['.blade.php', '.md'], '')
+            ->toString();
+    }
+
+    protected function getExtensionContent(string $relativePath, string $type): string
+    {
+        foreach (['.blade.php', '.md'] as $extension) {
+            $extensionPath = $this->customGuidelinePath($relativePath.'.'.$type.$extension);
+
+            if (file_exists($extensionPath)) {
+                return $this->renderBladeFile($extensionPath);
+            }
+        }
+
+        return '';
     }
 
     protected function getGuidelineAssist(): GuidelineAssist
--- a/src/Install/SkillWriter.php
+++ b/src/Install/SkillWriter.php
@@ -39,7 +39,7 @@ class SkillWriter
 
         $existed = is_dir($targetPath);
 
-        if (! $this->copyDirectory($skill->path, $targetPath)) {
+        if (! $this->copyDirectory($skill->path, $targetPath, $skill->custom ? null : $skill->name)) {
             return self::FAILED;
         }
 
@@ -123,7 +123,7 @@ class SkillWriter
         return @rmdir($path);
     }
 
-    protected function copyDirectory(string $source, string $target): bool
+    protected function copyDirectory(string $source, string $target, ?string $skillName = null): bool
     {
         if (! is_dir($source)) {
             return false;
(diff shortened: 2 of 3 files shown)


siarheipashkevich:
@calebdw it's what I needed!

But is there any way to exclude some guides from piece of the guidelines? For example, my app is api only and I don't need the rules: `## Frontend Bundling` `## Reading Browser Logs With the 'browser-logs' Tool` or `## Vite Error`

Or I use Passport instead of the Sanctum, and after updating boost I need to changed this
image

calebdw:
Glad to help out!

Interesting problem, that's really a separate concern though that does not apply to this PR though---might can consider in a future PR.

> But is there any way to exclude some guides from piece of the guidelines?

The first thing that comes to mind is maybe overriding the undesired guideline **with an empty file**? That should theoretically cause it to be excluded

> Or I use Passport instead of the Sanctum, and after updating boost I need to changed this

This is a different concern than extending or excluding---sounds like that guideline needs to be dynamic based on if Sanctum or Passport is used

siarheipashkevich:
@calebdw I got it, but anyway your PR will resolve some parts of my manual reviewing of the changes after the `boost:update` command, because I have some extra guides for `laravel` section.

> This is a different concern than extending or excluding---sounds like that guideline needs to be dynamic based on if Sanctum or Passport is used

It will be nice to have the solution for this or maybe write more general instruction for this line for avoiding direct mentions of the used tools

pushpak1300:
Thanks, I like the idea overall, but I have one concern. Most users don’t replace the default guidelines, and there’s already a workable way to customize them via the `.ai` folder. The downside is the added maintenance, especially since guideline order doesn’t actually affect behavior they’re all passed to the agent every time.

While this makes sense organisationally, I’m not convinced it’s worth adding right now. It introduces extra overhead and future maintenance for a problem few users currently face. Given that full or partial guideline overrides are already possible, I’d prefer to hold off until this becomes a more common issue.

calebdw:
@pushpak1300, I really need to extend the pest skill and give it conventions relevant to our project, and I would rather not have to duplicate the entire existing skill (which is good stuff) and I don't want to pollute the global guidelines with pest information that is not being used all the time (hence why it was moved to a skill in the first place).

I can update this PR to just focus on *skills* if you would be more willing to merge that?

pushpak1300:
@calebdw I’m not really against it but to better understand your use case could you please let me know?  You could simply copy the Boost generated skill to the `.ai/skills` folder and edit the `SKILL.md` file yourself to add your conventions as needed.

I understand you’ll miss some updated guidelines but if you’re familiar with the process you’ll be in a much better position. These guidelines are quite generic and if you can provide project-specific information to the app it will work really well..

calebdw:
> You could simply copy the Boost generated skill to the .ai/skills folder and edit the SKILL.md file yourself to add your conventions as needed.

Yeah that's a last resort, the biggest problem there is that now the maintenance burden for the core skill is on us instead of boost and we would need to pay closer attention to updates. Especially when the skill is updated for like a new major version

pushpak1300:
The Boost skills aren’t very opinionated, and I think adding testing instructions tailored to your project-specific setup would be more beneficial than relying solely on the out-of-the-box Boost skills. That said, I don’t think you need to worry about the update, since custom implementations tend to produce more accurate results than the default preset skills. I’m not opposed to the feature, but I don’t think it needs to be added at this point unless it becomes a real problem because of additional maintenance.

calebdw:
Should I close this then?

cswgr:
Thanks for implementing this!
But shouldn't this be documented here: https://laravel.com/docs/13.x/boost ?

calebdw:
in you missed it this was closed, not merged

cswgr:
Oh no, I thougt it was merged :(

PR #541, opened 2026-02-07
Enhance database-schema tool with full column metadata

### LINKED ISSUE #345: Database schema mcp lookup missing valuable information
### Laravel Package Version

1.7

### Laravel Version

12.37

### PHP Version

8.3

### System Info

Windows & Mac

### Description

The database schema mcp lookup is missing valuable information like:

  - Nullable/NOT NULL constraints
  - Default values
  - Character length limits
  - Precision for decimal types

These all exist when doing `php artisan db:table mydatabase.mytable`. 

### Steps To Reproduce

`php artisan db:table mydatabase.mytable` this commands shows all information required for the schema. 
Prompting the AI to check the same tables schema and then show output reveals the missing information.

### PHP ARTISAN
Image

### LARAVEL BOOST MCP
Image

Image

alanost:
Fixes #345

The `database-schema` MCP tool only returns an abstract type name per column (e.g. `string`, `integer`). AI agents are missing the information they need to generate correct validation rules, migrations, and model code: nullable constraints, default values, type precision, and auto-increment.

### What changed

Replaced `getColumnListing()` + `getColumnType()` with Laravel's `Schema::getColumns()`, which gives us full column metadata in a single call. Each column now includes:

- **type**: raw DB type with precision, e.g. `varchar(255)`, `decimal(8,2)` instead of just `string`
- **nullable**: whether the column allows NULL
- **default**: the column's default value
- **auto_increment**: whether the column auto-increments
- **comment** / **generation**: only included when actually set

Before:
```json
{ "id": { "type": "integer" }, "name": { "type": "string" } }
```

After:
```json
{
  "id": { "type": "integer", "nullable": false, "default": null, "auto_increment": true },
  "name": { "type": "varchar(255)", "nullable": false, "default": null, "auto_increment": false }
}
```

### Files changed (3 files, +30 -6)

- `src/Mcp/Tools/DatabaseSchema.php`: rewrote `getTableColumns()` to use `Schema::getColumns()`, updated tool description
- `tests/Feature/Mcp/Tools/DatabaseSchemaTest.php`: extended existing test with metadata assertions, added `DB::disconnect()` for Windows cleanup
- `.ai/boost/core.blade.php`: added a line pointing agents to the `database-schema` tool for table inspection

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema.php
+++ b/src/Mcp/Tools/DatabaseSchema.php
@@ -24,7 +24,7 @@ class DatabaseSchema extends Tool
     /**
      * The tool's description.
      */
-    protected string $description = 'Read the database schema for this application. Returns table names, columns (type with precision, nullable, default, auto_increment), indexes, and foreign keys. Use "filter" to narrow down tables by name (substring match).';
+    protected string $description = 'Read the database schema for this application. Returns table names, tables, columns (type only), indexes, foreign keys. Params: "database" - connection name; "filter" - substring match on table names; "include_column_details" (default false) - adds nullable, default, auto_increment, comments; "include_views" (default false); "include_routines" (default false) - stored procedures, functions, sequences.';
 
     /**
      * Get the tool's input schema.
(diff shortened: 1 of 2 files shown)


pushpak1300:
I changed this to be opt it because for larger db its filling up context much faster.

PR #548, opened 2026-02-09
feat: To add support for Cline

shenjian-8:
Support for Cline

### DIFF
--- a/src/Mcp/Tools/DatabaseQuery.php
+++ b/src/Mcp/Tools/DatabaseQuery.php
@@ -77,11 +77,51 @@ class DatabaseQuery extends Tool
         $connectionName = $request->get('database');
 
         try {
+            $connection = DB::connection($connectionName);
+            $prefix = $connection->getTablePrefix();
+
+            if ($prefix) {
+                $query = $this->addPrefixToQuery($query, $prefix);
+            }
+
             return Response::json(
-                DB::connection($connectionName)->select($query)
+                $connection->select($query)
             );
         } catch (Throwable $throwable) {
             return Response::error('Query failed: '.$throwable->getMessage());
         }
+    }
+
+    protected function addPrefixToQuery(string $query, string $prefix): string
+    {
+        $cteNames = $this->extractCteNames($query);
+
+        $pattern = '/\b(FROM|JOIN|INTO|UPDATE|TABLE|DESCRIBE|DESC)\s+([`"\']?)(\w+)\2/i';
+
+        return preg_replace_callback($pattern, function (array $matches) use ($prefix, $cteNames): string {
+            $keyword = $matches[1];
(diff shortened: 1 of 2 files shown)


src/Install/Agents/Cline.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function getMcpConfigPath(): string
+    {
+        //Ensure that there are no null entries in cline_mcp_settings.json
+        $home = getenv('HOME');
+        $platform = Platform::current();
+        return match ($platform) {
+            Platform::Darwin => $home.'/Library/Application Support/Code/User/globalStorage/saoudrizwan.claude-dev/settings/cline_mcp_settings.json',
+            Platform::Linux => $home.'/.config/Code/User/globalStorage/saoudrizwan.claude-dev/settings/cline_mcp_settings.json',
+            Platform::Windows => $home.'\AppData\Roaming\Code\User\globalStorage\saoudrizwan.claude-dev\settings\cline_mcp_settings.json',
```
pushpak1300, line 95: this paths should be global not your paths.

src/Install/Agents/Cline.php
```diff
@@ -0,0 +1,98 @@
...
+                'command' => 'command -v code',
+            ],
+            Platform::Windows => [
+                'paths' => [
+                    '%ProgramFiles%\\Microsoft VS Code',
+                    '%LOCALAPPDATA%\\Programs\\Microsoft VS Code',
+                ],
+            ],
+        };
+    }
```
pushpak1300, line 40: This is for vscode and not cline if user has installed vscode that does not mean user has cline we need specific detection for cline. there seems to be cline cli which could work in this case to detect automatically.

src/Install/Agents/Cline.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /** {@inheritDoc} */
+    public function mcpServerConfig(string $command, array $args = [], array $env = []): array
+    {
+        return [
+            'autoApprove' => [],
+            'timeout' => 300,
+            'type' => 'stdio',
+            'disabled' => false,
```
pushpak1300, line 69: how many of this fields are necessary fields ? if not we can just add necessary fields so server gets activated automatically.

pushpak1300:
Thanks for your pull request!

Unfortunately, i just check cline still does not support project level mcp so it doesn't make sense to add that at this point. https://github.com/cline/cline/discussions/2418


shenjian-8:
Thanks for the suggestion! I'll extract this functionality into a standalone Laravel package and publish it on Packagist. Once ready, I'll share the link here for anyone interested. This approach keeps the core project lean while still providing the feature to users who need it.

shenjian-8:
https://packagist.org/packages/laravel-boost/cline

Issue #540, opened 2026-02-07
Feature request: Google Antigravity


stdrew:
Could you consider adding it to the CLI / workflow automation?

It's becoming very popular and like Cursor it handles context files a little different than VS Code.

From docs (regarding skills): 
```
<workspace-root>/.agent/skills/<skill-folder>/	(Workspace)
~/.gemini/antigravity/skills/<skill-folder>/	(Global)
```
[Skills Docs](https://antigravity.google/docs/skills)

Thanks so much. Will use manual install/update for now.

pushpak1300:
MCP support is still not very solid with antigravity. I'm happy to add guidelines and skills support though. feel free to PR. Thanks.

pushpak1300:
https://github.com/laravel/boost/pull/554#issuecomment-3871493854

PR #554, opened 2026-02-09
feat(boost): add Antigravity agent with guidelines and skills support

soleinjast:
## What’s new

This PR introduces **Google Antigravity** as a supported agent in Laravel Boost.

It adds first-class support for:
- **Guidelines** (`AGENT.md`) with Antigravity-specific transformations
- **Workspace skills** via `.agent/skills`, aligned with Antigravity’s documented context model

This allows Boost users to integrate Antigravity into existing CLI and workflow automation in the same way as other supported agents.

---

## Implementation details

- Adds a new `Antigravity` agent implementation
- Supports:
  - Project guidelines (`AGENT.md`)
  - Workspace skills (`.agent/skills`)
- Maps MCP configuration to Antigravity’s global config location  
  (`~/.gemini/antigravity/mcp_config.json`) without attempting full MCP automation, as MCP support is still evolving for Antigravity
- Includes a small guideline transformation to handle Antigravity’s parsing behavior

---

## Backward compatibility

- No existing agents or workflows are affected
- No breaking changes

---

## Related discussion

Closes / relates to: #540

### DIFF
--- a/src/Install/Agents/Agent.php
+++ b/src/Install/Agents/Agent.php
@@ -228,11 +228,21 @@ abstract class Agent
     /**
      * Normalize command by splitting space-separated commands into command + args.
      *
+     * Absolute paths (starting with / on Unix or a drive letter on Windows)
+     * are never split, as they may contain spaces (e.g. macOS "Application Support").
+     *
      * @param  array<int, string>  $args
      * @return array{command: string, args: array<int, string>}
      */
     protected function normalizeCommand(string $command, array $args = []): array
     {
+        if (str_starts_with($command, '/') || preg_match('#^[a-zA-Z]:[/\\\\]#', $command)) {
+            return [
+                'command' => $command,
+                'args' => $args,
+            ];
+        }
+
         $parts = str($command)->explode(' ');
 
         return [
--- a/tests/Unit/Install/Agents/AgentTest.php
+++ b/tests/Unit/Install/Agents/AgentTest.php
@@ -380,6 +380,17 @@ test('splits valet php into command and arguments', function (): void {
     ]);
 });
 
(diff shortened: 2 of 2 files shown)


src/Install/Agents/Antigravity.php
```diff
@@ -0,0 +1,67 @@
...
+        return [
+            'paths' => ['.agent'],
+            'files' => ['AGENT.md'],
+        ];
+    }
+
+    public function mcpConfigPath(): string
+    {
+        return '~/.gemini/antigravity/mcp_config.json';
+    }
```
pushpak1300: Antigravity doesn't support project level mcp. can you please remvoe the support for MCP ?

soleinjast: right! Antigravity doesn’t support project-level MCP. I initially added it to reference Antigravity’s global MCP setup for consistency with other agents, but that was misleading in this context, so I’ve removed MCP support..

pushpak1300:
I’m a bit torn, but my leaning is: until AntiGravity supports project-level MCP, we probably shouldn’t add support on our side, since it could confuse end users.

That said, AntiGravity already supports the `.agent` standard, so if someone installs the `Opencode`, the AntiGravity skill and guideline support should get enabled automatically without extra setup.

Net: I’d prefer we hold off on merging this for now, at least until project-level MCP is supported.

stdrew:
@soleinjast 👊 thx for the additions even if on hold for now. 
@pushpak1300 new Laravel fan & sry for requesting w/o PR. good catch on shotty/non-workspace MCP. as it happens, the rush of excitement for AG is already cooling down…AI hype cycles are wild.

still a v nice idea…if Google decides to keep caring about it (coin flip ☺️)

PR #557, opened 2026-02-10
Simplify Tinker tool to wrap artisan tinker --execute

soleinjast:
## Summary

- Replaces PsySH-based `TinkerExecutor` with a thin `Artisan::call('tinker', ['--execute' => $code])` wrapper
- Follows the same pattern as `ListRoutes` — delegate to an Artisan command, capture output via `BufferedOutput`
- Deletes `TinkerExecutor.php` entirely (**-260 lines net**)

## Why

The `TinkerExecutor` manually bootstrapped PsySH (`Shell`, `Configuration`, `ClassAliasAutoloader`) — duplicating what `artisan tinker --execute` already does internally. By delegating to the Artisan command, we get correct Tinker behavior without maintaining PsySH integration code.

## What changed

| File | Change |
|------|--------|
| `src/Mcp/Tools/Tinker.php` | Rewritten to use `Artisan::call('tinker')` |
| `src/Mcp/Tools/Tinker/TinkerExecutor.php` | **Deleted** |
| `tests/Feature/Mcp/Tools/TinkerTest.php` | Simplified to minimal test coverage |
| `tests/Feature/Mcp/ToolExecutorTest.php` | Updated for text-based output |

## Tests

Minimal test coverage for the wrapper: happy path, error handling, PHP tag stripping, and environment eligibility.

### DIFF
--- a/tests/Feature/Mcp/Tools/TinkerTest.php
+++ b/tests/Feature/Mcp/Tools/TinkerTest.php
@@ -53,6 +53,15 @@ test('strips php tags from code', function (): void {
         ->toolTextContains('stripped');
 });
 
+test('executes code without semicolon', function (): void {
+    $tool = new Tinker;
+    $response = $tool->handle(new Request(['code' => 'echo 1']));
+
+    expect($response)->isToolResult()
+        ->toolHasNoError()
+        ->toolTextContains('1');
+});
+
 test('should register only in local environment', function (): void {
     $tool = new Tinker;
 


soleinjast:
Hey @pushpak1300 👋  

I’ve implemented the PsySH-based Tinker wrapper we discussed. Would appreciate your review when you get a chance 🙂

pushpak1300:
@soleinjast, is there a reason we’re not wrapping this in `php artisan tinket -e “echo $a”` and using the PSYSH execution engine by default?  There’ll be more edge cases with PsySH.

soleinjast:
> @soleinjast, is there a reason we’re not wrapping this in `php artisan tinket -e “echo $a”` and using the PSYSH execution engine by default? There’ll be more edge cases with PsySH.

Good question!

the main reason I initially used PsySH directly was to avoid double bootstrapping the console layer and to preserve structured results. Using PsySH directly allowed us to return both the evaluated result and captured output separately (including type and class information), whereas artisan tinker --execute only provides plain text output. 

That said, I completely understand the benefit of wrapping the official command for stricter alignment with Laravel’s console layer and reducing long-term maintenance risk. I’m happy with either direction — I mainly wanted to ensure we delegate to the real Tinker engine rather than reimplement execution semantics.

Issue #556, opened 2026-02-10
Boost install failing on windows 11


planetarquasar:
### Boost Version

2.1.1

### Laravel Version

12.47.0

### PHP Version

8.4

### System Info

Windows 11

### Description

The boost install failing.
php artisan boost:install

██████╗   ██████╗   ██████╗  ███████╗ ████████╗
██╔══██╗ ██╔═══██╗ ██╔═══██╗ ██╔════╝ ╚══██╔══╝
██████╔╝ ██║   ██║ ██║   ██║ ███████╗    ██║   
██╔══██╗ ██║   ██║ ██║   ██║ ╚════██║    ██║   
██████╔╝ ╚██████╔╝ ╚██████╔╝ ███████║    ██║   
╚═════╝   ╚═════╝   ╚═════╝  ╚══════╝    ╚═╝   

 ✦ Laravel Boost :: Install :: We Must Ship ✦ 

 Let's give MyApp a Boost

   Symfony\Component\Process\Exception\ProcessStartFailedException 

  The command "where claude 2>nul" failed.

Working directory: C:\FolderParent\FolderChild\FolderGrandchild

Error: proc_open(): CreateProcess failed: The parameter is incorrect

  at vendor\symfony\process\Process.php:382
    378▕             restore_error_handler();
    379▕         }
    380▕
    381▕         if (!$process) {
  ➜ 382▕             throw new ProcessStartFailedException($this, $lastError);
    383▕         }
    384▕         $this->process = $process;
    385▕         $this->status = self::STATUS_STARTED;
    386▕

  1   vendor\symfony\process\Process.php:252
      Symfony\Component\Process\Process::start()

  2   vendor\laravel\framework\src\Illuminate\Support\HigherOrderTapProxy.php:33
      Symfony\Component\Process\Process::run()

### Steps To Reproduce

composer require laravel/boost --dev
php artisan boost:install

soleinjast:
@pushpak1300 

Hi! The root cause is that `where claude 2>nul` uses shell redirection syntax which PHP's `proc_open` can't handle directly on Windows — it needs to go through `cmd.exe`.

Using `cmd /c where claude 2>nul` or `cmd /c where claude` instead of `where claude 2>nul` should fix it.

Also worth wrapping the detection in a `try/catch` so the installer fails gracefully instead of throwing a `ProcessStartFailedException` when Claude CLI is not installed.

Happy to submit a PR if that helps.

pushpak1300:
[...]

PR #558, opened 2026-02-10
Use cmd /c for Windows agent detection commands

soleinjast:
## Problem

On Windows, the agent detection commands (e.g. `where claude 2>nul`) use shell  redirection syntax which PHP's `proc_open` cannot handle directly — it needs to be routed through `cmd.exe`. This caused `boost:install` to crash with a 
`ProcessStartFailedException` instead of gracefully detecting whether the agent is installed.

## Fix

Prefixed all Windows detection commands with `cmd /c` across all affected agents:
- `ClaudeCode`
- `Codex`
- `Gemini`
- `OpenCode`

related #556


pushpak1300:
Thanks !

PR #561, opened 2026-02-10
Fix missing output key in Tinker tool error response

Orlando-Villanueva:
## Summary

The Tinker tool's error response (catch block) was missing the `output` key that the success response includes. This causes a `TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')` in the client when executed code throws an error.

```
⚙ laravel-boost_browser-logs [entries=10]
TypeError: undefined is not an object (evaluating 'output.output.toLowerCase')
```

## Changes

- Added `'output' => ob_get_contents() ?: ''` to the catch block's `Response::json()` call in `src/Mcp/Tools/Tinker.php`
- Added comprehensive tests verifying:
  - Error responses include the `output` key
  - Success and error responses have matching structure
  - Client-side operations (like `toLowerCase`) work safely on both response types

### DIFF
--- a/src/Mcp/Tools/Tinker.php
+++ b/src/Mcp/Tools/Tinker.php
@@ -72,6 +72,7 @@ class Tinker extends Tool
                 'type' => $throwable::class,
                 'file' => $throwable->getFile(),
                 'line' => $throwable->getLine(),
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks !

PR #568, opened 2026-02-12
Truncate large log entries in LastError tool response

leek:
## Summary

- Uses `Str::limit()` to cap log entries at 500 characters in `LastError::handle()` before returning file-based error entries via `Response::text()`

## Problem

Large stack traces in `laravel.log` (100+ lines per entry) can exceed MCP transport limits, causing the `last-error` tool to fail or return unusable payloads to AI agents.

## Test plan

- [ ] Trigger an error that produces a large stack trace (e.g. nested Livewire/Filament exception)
- [ ] Call the `last-error` MCP tool — should return a truncated entry (≤500 chars) with `...` suffix
- [ ] Trigger a small error (< 500 chars) — should return the full entry unchanged

### DIFF
--- a/.ai/folio/skill/folio-routing/SKILL.blade.php
+++ b/.ai/folio/skill/folio-routing/SKILL.blade.php
@@ -46,6 +46,27 @@ Always create new `folio` pages and routes using `{{ $assist->artisanCommand('fo
 
 // Creates: resources/views/pages/products/[id].blade.php → /products/{id}
 {{ $assist->artisanCommand('folio:page "products/[id]"') }}
+
+// Creates: resources/views/pages/users/[User].blade.php → /users/{user} (implicit model binding)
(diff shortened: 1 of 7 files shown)


src/Concerns/ReadsLogs.php
```diff
@@ -117,6 +117,24 @@ protected function isErrorEntry(string $line): bool
...
+
+        if (count($lines) <= $maxLines) {
+            return $entry;
+        }
+
+        $remaining = count($lines) - $maxLines;
+
+        return implode("\n", array_slice($lines, 0, $maxLines))
+            ."\n[... truncated {$remaining} more lines]";
+    }
```
pushpak1300: Instead of this can we not replace directly with the https://laravel.com/docs/11.x/strings#method-str-limit  with certain chars let's say 500 ?

pushpak1300:
Thanks !

PR #581, opened 2026-02-17
Support discovery for modular directory structures

soleinjast:
## Summary

This PR adds support for **modular Laravel project discovery** in Laravel Boost by consuming architecture signals exposed by `laravel/roster`.

By default, Boost only scans the `app/` directory. With this change, when Roster detects a **MODULAR** project structure, Boost will also discover classes located under a root-level `modules/` (or `Modules/`) directory.

---

## Motivation

Many Laravel applications use a modular architecture where domain code (such as Eloquent models, controllers, or enums) lives outside the `app/` directory.

Previously, Boost was unable to detect these classes, resulting in incomplete or misleading context during AI-assisted workflows.

This change delegates **architecture detection entirely to `laravel/roster`**, allowing Boost to remain simple and deterministic while supporting modular Laravel applications.

---

## Behavior

- Boost continues to scan the `app/` directory by default
- When Roster reports the `MODULAR` approach:
  - Boost additionally scans:
    - `modules/`
    - `Modules/` (case-insensitive)
    - `app-modules/`
- For all other project types, Boost falls back to the default behavior
- No filesystem heuristics or user configuration are required

---

## Related Work

- Project architecture detection is handled by `laravel/roster`
- Boost only consumes the detected approach and adjusts class discovery accordingly

related #465


calebdw:
This is something that whatever modular package you are using should provide---for example: https://github.com/InterNACHI/modular/tree/main/resources/boost

pushpak1300:
Thanks @calebdw this is really good with the new [option](https://github.com/laravel/boost/pull/580) we added managing modular will be much more easier with lot of customisation on the guidelines. 

I really like the approach if it's working really great i don't think we should manage this in boost.

soleinjast:
@pushpak1300 I’ve left a comment on #465  that’s related to this PR.

ItsReddi:
Hello,
any support needed here to get that merged?

pushpak1300:
can we fix the merge conflicts here ?

soleinjast:
> can we fix the merge conflicts here ?

@pushpak1300 It's ready!

PR #582, opened 2026-02-18
fix: blade skills with code before frontmatter are parsed correctly

calebdw:
Hello!

I was trying to make a 3rd party package skill dynamic, but it wasn't working because it wasn't passing the frontmatter test.

The fix is to render blade **before** trying to parse the frontmatter

Thanks!


pushpak1300:
Thanks.

PR #588, opened 2026-02-19
Fix read-only bypass in DatabaseQuery via CTE-wrapped writes

sulimanbenhalim:
Queries like `WITH x AS (SELECT 1) DELETE FROM users` bypass the read-only guard because the regex only checks that `SELECT` appears somewhere after `WITH`. The fix verifies the statement after the CTE is a `SELECT` and explicitly blocks write keywords after a closing paren.


pushpak1300:
Thanks !

PR #589, opened 2026-02-19
Allow overriding the browser log channel

SanderMuller:
`registerBrowserLogger()` unconditionally overwrites `logging.channels.browser` at runtime, making it impossible for users to customize the browser log channel in their `config/logging.php`.

This adds a null check so the default is only applied when the user hasn't defined their own `browser` channel.


pushpak1300:
Thanks !

PR #595, opened 2026-02-21
Add summary mode to DatabaseSchema tool

sulimanbenhalim:
Adds a `summary` parameter to `DatabaseSchema` that returns only table names with column types — ~83% fewer tokens. AI agents can call summary first, then drill down with `filter` on specific tables.

No breaking changes — defaults to `false`.

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema.php
+++ b/src/Mcp/Tools/DatabaseSchema.php
@@ -24,7 +24,7 @@ class DatabaseSchema extends Tool
     /**
      * The tool's description.
      */
-    protected string $description = 'Read the database schema for this application. Returns table names, tables, columns (type only), indexes, foreign keys. Params: "database" - connection name; "filter" - substring match on table names; "include_column_details" (default false) - adds nullable, default, auto_increment, comments; "include_views" (default false); "include_routines" (default false) - stored procedures, functions, sequences.';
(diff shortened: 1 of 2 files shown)


src/Mcp/Tools/DatabaseSchema.php
```diff
@@ -134,6 +146,28 @@ protected function getAllTablesStructure(?string $connection, string $filter = '
...
+        foreach ($this->getAllTables($connection) as $table) {
+            $tableName = is_object($table) ? $table->name : ($table['name'] ?? '');
+
+            if ($filter !== '' && ! str_contains(strtolower($tableName), strtolower($filter))) {
+                continue;
+            }
+
+            $tables[$tableName] = collect(Schema::connection($connection)->getColumns($tableName))
+                ->pluck('type', 'name')
+                ->all();
```
pushpak1300, line 165: From summary mode, I’ll remove the column as well; only the table name is fine.

sulimanbenhalim:
Hey, thanks for the review!

I think removing column types would make summary mode a bit too bare — the AI would just have table names with no idea what's inside them, so it'd have to immediately call the full schema anyway.

With column types included, the AI can see things like `user_id` in posts or spot pivot tables like `post_tag`, and then only drill into the 2-3 tables it actually needs via `filter`.

Here's the before/after for a single table:

**Before (current, every call):**
```json
"users": {
  "columns": {"id": {"type": "integer"}, "name": {"type": "varchar"}, "email": {"type": "varchar"}},
  "indexes": {"sqlite_autoindex_users_1": {"columns": ["email"], "type": null, "is_unique": true, "is_primary": false}},
  "foreign_keys": [],
  "triggers": [],
  "check_constraints": []
}
```

**After (summary mode):**
```json
"users": {"id": "integer", "name": "varchar", "email": "varchar"}
```

That's ~83% fewer tokens across a full schema while still being useful as a first pass.

pushpak1300:
Thanks.

PR #596, opened 2026-02-21
Fix third-party skills excluded when aiGuidelines is uninitialized

sulimanbenhalim:
`SkillComposer::getThirdPartySkills()` silently drops all third-party skills when `GuidelineConfig::$aiGuidelines` is uninitialized — `?? []` makes `in_array()` always false. `GuidelineComposer` already handles this correctly with an `isset()` guard; this adds the same guard to `SkillComposer`.


pushpak1300:
Thanks

PR #598, opened 2026-02-22
Add support for Ampcode

dringrayson:
## Summary

Adds [Amp (Ampcode)](https://ampcode.com) as a supported AI agent in Laravel Boost, enabling `boost:install` to detect and configure the Boost MCP server for Amp users.

## Motivation

[Amp](https://ampcode.com) is a frontier-focused coding agent built to stay at the cutting edge of each new model generation. It recently [deprecated its VS Code and Cursor extensions](https://ampcode.com/news/the-coding-agent-is-dead) to go all-in on its CLI/TUI, which connects to editors (VS Code, Neovim, JetBrains) but runs independently, including headless and in CI. Amp supports MCP natively and has a growing user base among developers who prefer terminal-first workflows.

Currently, Amp users must manually configure `.amp/settings.json` to use Laravel Boost. This PR brings Amp to parity with the other supported agents, so `boost:install` detects and configures it automatically.

## Implementation

Uses the `SHELL` installation strategy via `amp mcp add {key} --workspace -- "{command}" {args}`, which writes the MCP config to the workspace's `.amp/settings.json`.

### Agent details

| Property | Value |
|---|---|
| Config file | `.amp/settings.json` (managed by `amp mcp add --workspace`) |
| Config key | `amp.mcpServers` |
| Strategy | `SHELL` (with direct file fallback for env vars) |
| Guidelines | `AGENTS.md` |
| Skills | `.agents/skills` |
| System detection | `command -v amp` / `where amp`, `~/.amp`, `~/.config/amp` |
| Project detection | `.amp/` directory only |

### Design decisions

- **SHELL over FILE:** Amp's config key `amp.mcpServers` uses a dot-prefixed flat key. The `FileWriter` uses Laravel's `data_set()` which interprets dots as nesting, producing `{"amp": {"mcpServers": ...}}` instead of the correct `{"amp.mcpServers": ...}`. The shell strategy delegates to Amp's own CLI, avoiding this entirely.
- **Direct file fallback for env vars:** The Amp CLI does not support passing environment variables (no `-e` flag). When env vars are present (e.g. Herd MCP with `SITE_PATH`), `installMcp()` falls back to writing `.amp/settings.json` directly, bypassing both the shell command and `FileWriter`. The command is normalized before writing so compound executables like `"herd php"` are correctly split into `command` + `args`.
- **Quoted command in shell template:** The `{command}` placeholder is quoted to handle PHP executable paths containing spaces.
- **HTTP MCP support:** Overrides `installHttpMcp()` to use `amp mcp add {key} --workspace {url}` so Nightwatch and other remote MCP servers install correctly.
- **Project detection uses only `.amp/`:** `AGENTS.md` was excluded to avoid false positives, since multiple agents (Codex, OpenCode, Copilot) share that file.
- **System detection includes `~/.amp` and `~/.config/amp`:** Fallback paths for when the `amp` binary isn't on `$PATH` but the installation or config directory exists.

## Changes

- **New:** `src/Install/Agents/Amp.php`
- **Modified:** `src/BoostManager.php` (register agent)
- **New:** `tests/Unit/Install/Agents/AmpTest.php` (10 tests)
- **Modified:** `tests/Unit/BoostManagerTest.php`, `tests/Unit/Install/AgentsDetectorTest.php` (updated for new agent count)

## Testing

All 650 tests pass. No existing tests broken. New tests cover: name, display name, installation strategy, shell command quoting, guidelines path, skills path, project detection config, MCP installation with env vars (verifies command normalization and direct file write), MCP installation without env vars (verifies shell delegation), and HTTP MCP installation.

### DIFF
--- a/src/Install/Mcp/FileWriter.php
+++ b/src/Install/Mcp/FileWriter.php
@@ -241,13 +241,13 @@ class FileWriter
     {
         $braceCount = 1;
         $length = strlen($content);
-        $inString = false;
+        $stringQuote = null;
         $escaped = false;
 
         for ($i = $openBracePos + 1; $i < $length; $i++) {
             $char = $content[$i];
 
-            if (! $inString) {
+            if ($stringQuote === null) {
                 if ($char === '{') {
                     $braceCount++;
                 } elseif ($char === '}') {
@@ -256,12 +256,11 @@ class FileWriter
                     if ($braceCount === 0) {
                         return $i;
                     }
+                } elseif (($char === '"' || $char === "'") && ! $escaped) {
+                    $stringQuote = $char;
                 }
-            }
-
-            // Handle string detection (similar to hasUnquotedComments logic)
-            if ($char === '"' && ! $escaped) {
-                $inString = ! $inString;
+            } elseif ($char === $stringQuote && ! $escaped) {
+                $stringQuote = null;
             }
 
             $escaped = ($char === '\\' && ! $escaped);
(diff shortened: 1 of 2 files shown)


src/Install/Agents/Amp.php
```diff
@@ -0,0 +1,133 @@
...
+            return parent::installMcp($key, $command, $args, $env);
+        }
+
+        return $this->writeToSettingsFile($key, $command, $args, $env);
+    }
+
+    /** {@inheritDoc} */
+    public function installHttpMcp(string $key, string $url): bool
+    {
+        $result = Process::run("amp mcp add {$key} --workspace {$url}");
```
pushpak1300: you have to always use file mode to correctly add support. 

If you run php artisan boost:install it doesn't ovewrite the config.

dringrayson: Totally agree with your feedback, thank you for catching this.

I updated the implementation to always use file-based writes for Amp so `boost:install` correctly updates existing MCP entries in `.amp/settings.json` on reruns. This now writes directly to `amp.mcpServers` for both local command MCPs and HTTP MCPs, instead of relying on `amp mcp add` behavior.

I also added coverage to confirm overwrite behavior and preservation of unrelated Amp settings, plus invalid JSON handling.

Fixed in f64aaf7

PR #604, opened 2026-02-25
Fix SearchDocs crashing when MCP client passes array params as JSON strings

digitall-it:
## Problem

Some MCP clients (e.g. Claude Code when a tool is deferred and not yet loaded) serialize array parameters as JSON-encoded strings instead of native PHP arrays. This causes `SearchDocs::handle()` to throw a fatal error:

```
array_map(): Argument #2 ($array) must be of type array, string given
```

This affects both the `queries` and `packages` parameters.

## Solution

Decode the parameter with `json_decode()` when it arrives as a string, falling back gracefully to an empty array (`queries`) or `null` (`packages`).

## Tests

Added two new tests to `SearchDocsTest.php`:
- `it handles queries passed as a JSON-encoded string`
- `it handles packages passed as a JSON-encoded string`

### DIFF
--- a/src/Mcp/Tools/SearchDocs.php
+++ b/src/Mcp/Tools/SearchDocs.php
@@ -55,28 +55,17 @@ class SearchDocs extends Tool
     public function handle(Request $request): Response|Generator
     {
         $apiUrl = config('boost.hosted.api_url', 'https://boost.laravel.com').'/api/docs';
-        $packagesFilterRaw = $request->get('packages');
 
-        if (is_string($packagesFilterRaw)) {
-            $packagesFilter = json_decode($packagesFilterRaw, true);
+        $packagesFilter = $this->resolveArrayParam($request->get('packages'));
 
-            if (json_last_error() !== JSON_ERROR_NONE || ! is_array($packagesFilter)) {
-                return Response::error('Invalid packages parameter: '.json_last_error_msg());
-            }
-        } else {
-            $packagesFilter = $packagesFilterRaw;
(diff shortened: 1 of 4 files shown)


pushpak1300:
@digitall-it can you share more info on how ai calling the tool some examples ?

digitall-it:
Hi @pushpak1300\! Happy to explain.

The issue is triggered by Claude Code's [MCP Tool Search](https://code.claude.com/docs/en/mcp#scale-with-mcp-tool-search) feature.

When many MCP servers are configured, Claude Code automatically **defers tool definitions** to save context window space — they are **not loaded upfront**. Instead, Claude calls a `ToolSearch` tool at runtime to **discover and load only the tools it actually needs**.

The problem occurs when Claude attempts to call a deferred tool **before** loading it via `ToolSearch`. At that point, Claude **doesn't know the tool's JSON schema** (i.e. that `queries` and `packages` are `array` types), so the MCP client **serializes them as JSON-encoded strings** instead of native arrays.

For example, instead of passing:
```json
{ "queries": ["livewire wire:submit"], "packages": ["livewire/livewire"] }
```

It passes:
```json
{ "queries": "[\"livewire wire:submit\"]", "packages": "[\"livewire/livewire\"]" }
```

This causes `array_map()` to throw immediately since it receives a **`string` instead of an `array`**.

Once `ToolSearch` loads the tool and **the schema is known**, parameters are passed correctly. The defensive `json_decode` we added **handles the string case gracefully** in the meantime.

PR #605, opened 2026-02-26
Add support for filtering browser logs by timestamp

alihamze:
# Problem

You tell your agent that something it implemented isn't working and it goes to look at the browser logs to see what's happening. However the logs may contain older errors from prior, unrelated, issues. The agent pulls them all in and gets sidetracked chasing an error that isn't relevant anymore.

# Solution

Added a `since` parameter to the `BrowserLogs` tool. This lets the agent request only logs that occurred since a given timestamp. It uses `Carbon::parse` so it supports any format supported by Carbon including relative times.

Added a test case that uses `since` to limit the results returned by the tool.

## Backwards Compatible

If you don't send that parameter, the tool functions exactly as it did prior to this change. 

While I'd like to have a reasonable default value set for `since`, I think that'd be considered a breaking change. Let me know if not.


pushpak1300:
Thanks for your pull request!

PR #609, opened 2026-02-27
Add Adaptive Context Engine (ACE)

### LINKED ISSUE #600: Idea: Reduce MCP round-trips by batching context into fewer tools
Right now Boost registers 15 MCP tools. That means ~2,100 tokens of schemas loaded into every single message in the conversation, and the AI typically needs 3-5 sequential tool calls to gather context for a task (e.g. schema, routes, errors, app info one by one).

What if instead of 15 individual tools, there were just 2-3 that let the AI request everything it needs in one call?

Something like:
- A lightweight **manifest** tool that returns a compact catalog of whats available (~200 tokens instead of ~2,100)
- A **resolve** tool that accepts multiple context requests at once and returns them batched in a single respone
- Keep `execute` as-is for code execution

This could also open the door for pre-defined **bundles** like "give me everything I need for debugging" or "give me database context" so the AI doesn't even have to think about which pieces to request.

From some rough estimates this could cut schema overhead by ~80% and reduce round-trips from 5 to 2 for most tasks.

Happy to start working on this if you think it's worth exploring.

sulimanbenhalim:
## Summary

15 MCP tools → 3. Schema overhead drops ~80%, round trips go from 5 to 2 for most tasks.

`boost-manifest` gives the AI a compact catalog of whats available (~200 tokens instead of ~2,100). `resolve-context` batches multiple data sources into a single response. `execute` handles code execution.

Pre-built bundles like `@debug` and `@database-work` let the AI grab everthing it needs in one shot instead of figuring out which pieces to request individually. Read-only tools run in-process when ACE is on, so theres no subprocess overhead for context resolution.

Fully opt in. Set `BOOST_ACE_ENABLED=true` and thats it. Legacy tools can stay alongside during migration.

## How it works

The AI calls `boost-manifest` once to see a compact menu of available context slices and bundles. Then it calls `resolve-context` with whatever it needs, all in one call. Slices map directly to the existing tools (DatabaseSchema, ListRoutes, LastError, etc) so the actual data is identical, just delivered smarter.

Bundles are pre-defined combos: `@debug` expands to last-error + browser-logs + app-info. `@database-work` expands to db-schema + db-connections + app-info + foundation guidelines. Custom bundles can be registered and unwanted ones excluded via config.

Guidelines (foundation, php, boost, etc) are also available as slices so the AI can pull specific guidelines on demand instead of loading all of them upfront.

## Test plan

78 tests, 298 assertions covering registries, resolvers, tool execution, bundle expansion, deduplication, error isolation, and server boot with ACE on and off.

Closes #600

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -48,4 +48,28 @@ return [
         'vendor_bin' => env('BOOST_VENDOR_BIN_EXECUTABLE_PATH'),
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Adaptive Context Engine (ACE)
+    |--------------------------------------------------------------------------
+    |
+    | ACE consolidates MCP tools and guidelines into a compact manifest
+    | with batched context resolution. Skills remain delivered via
+    | .claude/skills/ files and are not affected by ACE.
+    |
+    | When legacy_tools is false (default), ACE replaces the 15 legacy
+    | tools with a 3-tool interface (manifest, resolve, execute) for
+    | significant token reduction. Set legacy_tools to true to keep
+    | legacy tools alongside ACE during a gradual migration.
+    |
+    */
+
+    'ace' => [
+        'enabled' => env('BOOST_ACE_ENABLED', false),
+        'legacy_tools' => false,
+        'bundles' => [
+            'exclude' => [],
+        ],
+    ],
+
 ];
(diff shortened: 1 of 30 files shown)


pushpak1300:
Closing in favour of #629

PR #616, opened 2026-03-02
Add security audit to add-skill command

pushpak1300:
The `boost:add-skill` command downloads skills from remote GitHub repos but has no security assessment of the source. The Laravel Skill Cloud audit API includes a non-blocking audit via partner services (ATH, Socket, Snyk) — this brings the same capability to Boost.

### Usage

```bash
# Install with audit (default)
php artisan boost:add-skill owner/repo --all

# Skip audit
php artisan boost:add-skill owner/repo --all --skip-audit
```

### Approach

- Calls the audit API at `skills.laravel.cloud/api/v1/skills/audit` after skill installation with a 3-second timeout
- Displays results as a color-coded table (red for critical/high, yellow for medium, green for safe/low)
- Non-blocking: audit failures never prevent installation and silently return empty results
- Adds `--skip-audit` flag to opt out entirely

Screenshot 2026-03-16 at 7 06 09 PM

### DIFF
--- a/src/Skills/Remote/Risk.php
+++ b/src/Skills/Remote/Risk.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Boost\Skills\Remote;
+
+enum Risk: string
+{
+    case Critical = 'critical';
+    case High = 'high';
+    case Medium = 'medium';
+    case Low = 'low';
+    case Safe = 'safe';
+
+    public function weight(): int
+    {
+        return match ($this) {
+            self::Critical => 5,
+            self::High => 4,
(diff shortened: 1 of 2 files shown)


src/Skills/Remote/AuditResult.php
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+namespace Laravel\Boost\Skills\Remote;
+
+class AuditResult
+{
+    public function __construct(
+        public string $partner,
+        public string $risk,
```
joetannenbaum: Maybe a `Risk` enum? Encapsulate weight, label, color under the enum?

PR #621, opened 2026-03-02
fix: register installed skills in boost.json during add-skill

meirdick:
## Summary

- `boost:add-skill` now registers installed skill names in `boost.json`'s `"skills"` array before calling `boost:update`
- This ensures the subsequent sync to agent directories (`.claude/skills/`, `.cursor/skills/`, etc.) always runs, even if the user never previously ran `boost:install --skills`

## Problem

When a user runs `boost:add-skill` for the first time (without having previously enabled skills via `boost:install --skills`), skills download to `.ai/skills/` but never sync to agent directories. The skill is invisible to the agent.

**Root cause:** `boost:update` checks `Config::hasSkills()` → reads `"skills"` key from `boost.json` → key doesn't exist → `hasSkills()` returns `false` → sync skipped.

Discovered while installing [meirdick/growth-product-context](https://github.com/meirdick/growth-product-context) — a community skill for product analytics journey discovery.

## Fix

Added `registerInstalledSkills()` which merges newly installed skill names into `boost.json`'s `"skills"` array before calling `runBoostUpdate()`. This ensures `Config::hasSkills()` returns `true` and the sync proceeds.

## Test plan

- [ ] Run `boost:install` without `--skills` on a fresh project
- [ ] Verify `boost.json` has no `"skills"` key
- [ ] Run `boost:add-skill meirdick/growth-product-context`
- [ ] Verify `boost.json` now has `"skills": ["growth-product-context"]`
- [ ] Verify `.claude/skills/growth-product-context/SKILL.md` exists
- [ ] Run `boost:add-skill` again with a second skill — verify both are in `boost.json`

### DIFF
--- a/src/Console/AddSkillCommand.php
+++ b/src/Console/AddSkillCommand.php
@@ -12,7 +12,6 @@ use Illuminate\Support\Facades\File;
 use InvalidArgumentException;
 use Laravel\Boost\Concerns\DisplayHelper;
 use Laravel\Boost\Skills\Remote\GitHubRepository;
-use Laravel\Boost\Support\Config;
 use Laravel\Boost\Skills\Remote\GitHubSkillProvider;
 use Laravel\Boost\Skills\Remote\RemoteSkill;
 use Laravel\Prompts\Terminal;
@@ -168,7 +167,6 @@ class AddSkillCommand extends Command
 
             grid($results['installedNames']);
 
-            $this->registerInstalledSkills($results['installedNames']);
             $this->runBoostUpdate();
             $this->showOutro();
         }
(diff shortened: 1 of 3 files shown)


pushpak1300:
Okay so if I understood this only happens when you try to install skills without running `php artisan boost:install --skills` ?  but why you would anyone not want to install boost skill and then download custom skills ?  

I'll need to rethink on this one if this is something we should support or not.

meirdick:
@pushpak1300 The issue that the added external skills are not recognized by the default Boost installer when running Boost:install. its highly confusing, when the command (boost:install) isnt picking up on the previous skill added and its expected by the user to be in the install list

meirdick:
used this today and was seamless! thank you.

Rami-Kassouf-FOO:
Are you sure this was merged and tested correctly? 

I had to do boost:install --skills for it to work

I didn't know about --skills and i installed boost and then added skills and it just went to .ai

meirdick:
Tested and verified. What agent/harness are you using and is it installed in the project folder?

PR #624, opened 2026-03-03
Add 'json' option for detailed output in ListRoutes

TheDoplarEffect:
## The Issue 
Current output truncates routes list which can hide output from the LLM such as middlewares or full file names for Controllers leading to LLM confusion.

## Solution

Expose the --json flag from `php artisan route:list` so that LLM can receive full output in json format


pushpak1300:
Thanks for your pull request!

PR #628, opened 2026-03-04
Add laravel-best-practices skill

pushpak1300:
Adds a comprehensive best practices skill with 125+ rules across 21 categories (database performance, security, caching, eloquent, validation, configuration, testing, deployment, queues, routing, HTTP client, events, mail, error handling, scheduling, architecture, migrations, collections, blade views, conventions, and advanced query patterns).

### DIFF
--- a/.ai/laravel/skill/laravel-best-practices/SKILL.md
+++ b/.ai/laravel/skill/laravel-best-practices/SKILL.md
@@ -67,6 +67,7 @@ Check sibling files, related controllers, models, or tests for established patte
 - Attribute casts in the `casts()` method
 - Cast date columns, use Carbon instances in templates
 - `whereBelongsTo($model)` for cleaner queries
+- Never hardcode table names — use `(new Model)->getTable()` or Eloquent queries
 
 ### 6. Validation & Forms → `rules/validation.md`
 
--- a/.ai/laravel/skill/laravel-best-practices/rules/eloquent.md
+++ b/.ai/laravel/skill/laravel-best-practices/rules/eloquent.md
@@ -120,3 +120,29 @@ Correct:
 Post::whereBelongsTo($user)->get();
 Post::whereBelongsTo($user, 'author')->get();
 ```
+
+## Avoid Hardcoded Table Names in Queries
+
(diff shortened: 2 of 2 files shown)


.ai/laravel/12/skill/laravel-best-practices/rules/advanced-queries.md
```diff
@@ -0,0 +1,106 @@
...
+Incorrect (correlated EXISTS re-executes per row):
+
+```php
+$query->whereHas('company', fn ($q) => $q->where('name', 'like', $term));
+```
+
+Correct (index-friendly lookup):
+
+```php
+$query->whereIn('company_id', Company::where('name', 'like', $term)->pluck('id'));
```
siarheipashkevich: @pushpak1300 `Company::where('name', 'like', $term)->pluck('id')` it will download all the found data, which can be a lot. This is not a good practice. Maybe better: `Company::where('name', 'like', $term)->select('id')` in that case we will get subquery in the sql where condition. But this is the same as above, just a different approach.

andrey-helldar:
The architecture guidelines set specific conditions for writing code that completely contradict the rules adopted by our team.

How do I prevent this skill from being installed?

I've tried the following:

- Removing it from the `boost.json` file and running `boost:update` - the skill is added and installed.
- Creating the `.ai/skills/laravel-best-practices/SKILL.md` file - the file is ignored.

pushpak1300:
@andrey-helldar Did you tried this because we specifically added this point.

> Before applying any rule, check what the application already does. Laravel offers multiple valid approaches — the best choice is the one the codebase already uses, even if another pattern would be theoretically better. Inconsistency is worse than a suboptimal pattern.
> 
> Check sibling files, related controllers, models, or tests for established patterns. If one exists, follow it — don't introduce a second way. These rules are defaults for when no pattern exists yet, not overrides.

andrey-helldar:
@pushpak1300 I haven't had a chance to try the new rules in practice yet. I saw the release and read the changes.

We have several legacy projects with very broken architectures, so we clearly don't define architecture skills—it would take several months.

For this reason, I'm considering disabling the skill, as I understand that since the AI ​​agent doesn't have any other architecture rules, it will obviously use these.

For now, I've already implemented a stub in all projects that prohibits the agent from using certain rules from the `laravel-best-practices` skill, and I've provided links to them.

But this is a workaround.

PR #645, opened 2026-03-10
Remove redundant Pennant guidelines

pushpak1300:
When we first added skill support, skill triggering wasn't reliable enough, so we added a `core.blade.php` guideline that duplicated instructions already in the Pennant skill — things like "always use `search-docs`" and "activate `pennant-development` for Pennant tasks". This burned extra context tokens on every request.

We now have an eval system for testing and improving skill descriptions, so they trigger reliably. This means the redundant guideline file is no longer needed.

### Approach

- Remove the redundant `core.blade.php` guideline that duplicated the Pennant skill's own instructions
- Refine the skill description to be more specific about Laravel Pennant scoping, validated through trigger evals


pushpak1300:
This is the result of the benchmarks.

| Aspect | Original Description | Final Description |
|--------|----------------------|-------------------|
| **Format** | Activates when... / or when... | Use when... Trigger whenever... Common tasks... Also trigger... Do not trigger... |
| **Test Score** | 8/20 (40% pass rate) | ~16/20 (80% pass rate) |
| **Key Improvement** | Failed to trigger on legitimate Pennant queries | Explicitly mentions Pennant by name + expanded common tasks |
| **Length** | Shorter, concise | Longer, more comprehensive |
| **Trigger Keywords** | "feature flags", "feature toggles", "@feature directive" | "Pennant" (by name) + all previous + class-based features, middleware, events, custom drivers, canary releases, dark launches |
| **Negative Guidance** | None | Explicitly excludes generic config, auth policies, authentication |

PR #651, opened 2026-03-11
Add `--discover` flag to `boost:update`

MrPunyapal:
## Summary
Introduces an opt-in discovery mode for `boost:update` via the `--discover` flag. Discovery now detects both new guideline packages and new skills and prompts the user to select which to add.

image

### DIFF
--- a/src/Console/UpdateCommand.php
+++ b/src/Console/UpdateCommand.php
@@ -27,7 +27,7 @@ class UpdateCommand extends Command
             return self::FAILURE;
         }
 
-        if ($this->shouldDiscover()) {
+        if ($this->option('discover')) {
             $this->discoverNewContent($config);
         }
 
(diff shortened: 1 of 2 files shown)


pushpak1300:
How does this solve this behaviour for example, if A B and C packages are available I need skills for package A so I choose the guidelines for A and ignore B and C.  Next time I install any other package even though it doesn’t require skills will it prompt me for installation every time?

MrPunyapal:
@pushpak1300 I made it optional now by adding `--discover`,  so it will only ask when we add the flag, else it will ignore.

MrPunyapal:
@pushpak1300 I rebased and fixed the failing test.

pushpak1300:
### Changes in `30185a4` — Remove per-skill allowlist and filtering from UpdateCommand

### Why
Discovery is now scoped to **packages only** (not individual skills). Per-skill filtering added unnecessary complexity — when a package is selected, all its skills are included automatically.

PR #657, opened 2026-03-12
Use major.minor PHP version in guidelines instead of full version

oddvalue:
## Description

This PR updates the PHP version output in generated guidelines and the `ApplicationInfo` MCP tool to use only the major and minor version (e.g. `8.4`) instead of the full version including the patch number (e.g. `8.4.3`).

## Motivation

On teams where different members may be running slightly different patch versions of PHP (e.g. `8.4.1` vs `8.4.3`), the full `PHP_VERSION` constant causes unnecessary churn in guideline files. Every time a team member with a different patch version runs the install command, the guideline file changes, leading to noisy diffs and constant re-generation. Since the patch version is rarely relevant to coding guidelines or API compatibility, using only the major and minor version keeps guideline files stable across the team.

## Changes

- **`.ai/foundation.blade.php`** — Changed `{{ PHP_VERSION }}` to `{{ PHP_MAJOR_VERSION }}.{{ PHP_MINOR_VERSION }}`
- **`src/Mcp/Tools/ApplicationInfo.php`** — Changed `PHP_VERSION` to `PHP_MAJOR_VERSION.'.'.PHP_MINOR_VERSION`
- **`tests/Feature/Mcp/Tools/ApplicationInfoTest.php`** — Updated test assertions to match the new format

## Impact

- No breaking changes — the version string format changes from `8.x.x` to `8.x`, which is still valid and more appropriate for guidelines
- This is consistent with how package versions are already displayed in the foundation template (using `$package->majorVersion()`)
- All existing tests pass

### DIFF
--- a/.ai/foundation.blade.php
+++ b/.ai/foundation.blade.php
@@ -8,7 +8,7 @@ The Laravel Boost guidelines are specifically curated by Laravel maintainers for
 ## Foundational Context
 This application is a Laravel application and its main Laravel ecosystems package & versions are below. You are an expert with them all. Ensure you abide by these specific packages & versions.
 
-- php - {{ PHP_VERSION }}
+- php - {{ PHP_MAJOR_VERSION }}.{{ PHP_MINOR_VERSION }}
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks.

PR #664, opened 2026-03-12
Configure sail binary path

iWader:
Currently the sail binary path is hard coded and cannot be customized. Due to legacy reasons we maintain a fork of sail which places the binary outside the `vendor/bin` directory.

Customizing this path allows for `boost:install` to correctly identify that sail is being used, and `boost:update` outputs the expected path in `AGENTS.md`, etc.

The sail executable path is intentionally not documented in the default published config. Since this is a niche option intended for non-standard Sail setups, users who need it can manually add `'sail' => env('BOOST_SAIL_EXECUTABLE_PATH')` to their `executable_paths` config. The `binaryPath()` method falls back to `DEFAULT_BINARY_PATH` when the key is absent, so there is no impact on standard installations.

### DIFF
--- a/.ai/foundation.blade.php
+++ b/.ai/foundation.blade.php
@@ -8,7 +8,7 @@ The Laravel Boost guidelines are specifically curated by Laravel maintainers for
 ## Foundational Context
 This application is a Laravel application and its main Laravel ecosystems package & versions are below. You are an expert with them all. Ensure you abide by these specific packages & versions.
 
-- php - {{ PHP_VERSION }}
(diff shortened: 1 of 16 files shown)


pushpak1300:
the sail executable path is intentionally not documented in the default published config. Since this is a niche option intended for non-standard Sail setups, users who need it can manually add `'sail' => env('BOOST_SAIL_EXECUTABLE_PATH')` to their `executable_paths` config.

iWader:
Thank you for fixing these up and merging :)

PR #666, opened 2026-03-14
Remove Herd MCP integration in favor of Herd CLI

pushpak1300:
Herd 1.26.0 introduced CLI commands that replace the functionality previously provided by the Herd MCP server. The `herd` CLI now supports service management, PHP version management, site listing, secure/unsecure, and isolate/unisolate — making the separate MCP integration unnecessary.

### Approach

- Removed the Herd MCP installation prompt, config, and writer logic from the install command
- Removed `isMcpAvailable()` and `mcpPath()` from the Herd class (kept `isInstalled()` for guideline/prompt use)
- Updated the Herd guideline to document the equivalent `herd` CLI commands
- Removed related tests and added a sail + nightwatch combo test

> [!WARNING]
> The `herd_mcp` config key in `boost.json` is no longer read or written. Existing projects with this key set will simply ignore it on next install.

### DIFF
--- a/.ai/herd/core.blade.php
+++ b/.ai/herd/core.blade.php
@@ -1,12 +1,4 @@
 # Laravel Herd
 
-- The application is served by Laravel Herd and will be available at: `https?://[kebab-case-project-dir].test`. Use the `get-absolute-url` tool to generate valid URLs for the user.
-- You must not run any commands to make the site available via HTTP(S). It is always available through Laravel Herd.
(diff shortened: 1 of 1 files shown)


pushpak1300:
> Looks good overall. My only minor concern is that we're removing and altering public methods, but I don't think anyone is likely to be using Boost programmatically, right? 

I guess so. even if they are using they might have probably pin the boost version in that case.

PR #669, opened 2026-03-16
Remove redundant "When to Apply" sections from skill bodies

sulimanbenhalim:
## What

Removes the `## When to Apply` section from 19 skill files across all variants (Livewire 2/3/4, Pest 3/4, Inertia React/Vue/Svelte v1/v2, Tailwind v3/v4, Folio, Flux UI free/pro, Volt, Wayfinder, Pennant)

## Why

Both Anthropic's published docs and their internal skill-creator skill say the same thing

The [Skill authoring best practices](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) are clear that the `description` field owns "when to use":

> "The description field enables Skill discovery and should include both what the Skill does and **when to use it**."

It also asks authors to challenge every token: "Does Claude really need this explanation?"

The skill-creator skill (Anthropic's meta-skill for building skills) goes further:

> "All 'when to use' info goes here [in the description], not in the body."

Every removed section was already covered by the skill's `description` frontmatter — nothing was lost, just duplicate content loaded into context on every trigger

## Verification

Each skill's `## When to Apply` bullets were checked against its `description` field before removal — every bullet was already there. No triggering behavior changes, since the description is the only thing Claude reads to decide whether to invoke a skill

### DIFF
--- a/.ai/fluxui-free/skill/fluxui-development/SKILL.blade.php
+++ b/.ai/fluxui-free/skill/fluxui-development/SKILL.blade.php
@@ -10,14 +10,6 @@ metadata:
 @endphp
 # Flux UI Development
 
-## When to Apply
-
-Activate this skill when:
-
-- Creating UI components or pages
-- Working with forms, modals, or interactive elements
-- Checking available Flux components
-
 ## Documentation
 
 Use `search-docs` for detailed Flux UI patterns and documentation.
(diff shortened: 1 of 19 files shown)


pushpak1300:
Thanks this make sense. The earlier skill-creator plugin add those section but the updated one is removing them so make sense to remove it. Also this needs to be removed from individual skill.

PR #670, opened 2026-03-16
Update pest-testing skill descriptions and clean up core guidelines

pushpak1300:
The pest-testing skill descriptions were too broad and inconsistent,
causing unreliable activation. The core guidelines also had redundant
directives that are now covered by the improved skill descriptions.


pushpak1300:
| Old | New | Delta |
|-----|-----|-------|
| 10/20 (50%) | 14/20 (70%) | +20% overall, +40% recall, 0% precision change |

PR #700, opened 2026-03-18
Support for third party MCP servers to Laravel Boost

### LINKED ISSUE #522: Feature proposal: Suport third party mcp server configuration definition like for skills and guidelines
Hi probably is a bit out of focus for boost but as now it supports third party packages skills and guidelines, would it be possible to add support for third party mcp server definition. 

This could check into the packages for a predefined location file and the json content from there would be merged with the laravel boost server definition.

Right now third party packages exposes guidelines under 
- resources/boost/guidelines/core.blade.php
- resources/boost/skills/{skill-name}/SKILL.md
 
I propose to add a third one
- resources/boost/mcp/mcp.json
 tis will check the json inside that file, retrieve the "servers" array and merge the contents with the boost mcp server definition before writing it to the specific agent supported folder. 

If this is something inline with the project I am available to create a pr for it.

## Context
I want to expose the laraplugins.io mcp server as a packagist package that provides skills, guidelines and the mcp server configuration to agents. right now the mcp server must be installed manually but this could be integrated into boost I think. Alos Imagine this to be useful for other packages in the long term too.

wilsenhc:
## Adds Third Party MCP support to Laravel Boost

image

I have also created this small demo package to test this feature works: `composer require wilsenhc/laravel-boost-third-party-mcp-test`.

Closes #522

MCPs are not dead yet 🚀

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -290,7 +290,8 @@ class InstallCommand extends Command
      */
     protected function selectThirdPartyPackages(): Collection
     {
-        $packages = ThirdPartyPackage::discover();
+        $packages = ThirdPartyPackage::discover()
+            ->filter(fn (ThirdPartyPackage $pkg): bool => $pkg->hasGuidelines || $pkg->hasSkills);
 
         if ($packages->isEmpty()) {
             return collect();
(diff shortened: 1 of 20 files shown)


wilsenhc:
@pushpak1300 I'll close this PR for now as it is very out of date.

I will re-visit the current implementation of this feature and come up with a better PR in the future. 

Some feedback would be appreciated too if you have an opinion of the direction this should feature should take.

pushpak1300:
@wilsenhc Thanks for all the work. I discussed this internally but it hasn’t proven a real use case yet.  No one’s keen on adding mcp serve via composer except a few packages.  While I’m interested, I’m actually on fence about it.

PR #722, opened 2026-03-26
Conditionally render MCP guideline sections

### LINKED ISSUE #721: MCP references exist when MCP server is not installed
Hello,

My company primarily uses GitHub Coding Agent and we are not currently allowed to use MCP servers. As a result I had tried configuring Boost without the MCP server, i.e. installing like `php artisan boost:install -n --guidelines --skills` (leaving off `--mcp`). While this leaves a lot of enhancements on the table, it means we still get to use the excellent guidelines and skills provided by Boost.

However, this still inserts references to MCP tools in AGENTS.md:

># Laravel Boost
>
>## Tools
>
>- Laravel Boost is an MCP server with tools designed specifically for this application. Prefer Boost tools over manual alternatives like shell commands or file reads.
>- Use `database-query` to run read-only queries against the database instead of writing raw SQL in tinker.
>- Use `database-schema` to inspect table structure before writing migrations or models.
>- Use `get-absolute-url` to resolve the correct scheme, domain, and port for project URLs. Always use this before sharing a URL with the user.
>- Use `browser-logs` to read browser logs, errors, and exceptions. Only recent logs are useful, ignore old entries.

It would be nice if, when the MCP option is not used during install, the references to the MCP server and tooling is omitted from the generated agent configuration.

Xiol:
Before this change, the Tools and Searching Documentation sections were always emitted into the generated AI guidelines, even when users chose not to enable MCP during installation. This meant AI agents were instructed to use database-query, search-docs, and other MCP tools that simply weren't available in their environment, leading to failed tool calls or confusing behaviour.

After this change, those sections are only included when MCP was actually selected at install time. Users who run a guidelines-only setup get a leaner, accurate guideline file that reflects their actual configuration.

Closes #721.


Xiol:
Thanks for your review, good catch there! I'm now passing the `--mcp` flag from UpdateCommand to InstallCommand during boost:update, so the MCP sections are retained.

pushpak1300:
But MCP config once published isn't supposed to change. it will be kind of breaking change,

Xiol:
How about this? I've updated `buildGuidelineConfig()` to fall back to the stored config value when MCP isn't in the selected features. This way the guideline sections are preserved during updates without re-triggering MCP installation.

PR #729, opened 2026-03-27
Add mcp_config_path configuration for monorepo support

### LINKED ISSUE #709: Add mcp_config_path configuration for monorepo support
## Problem

In monorepo setups where the Laravel application lives in a subdirectory (e.g., `project-root/laravel/`), the `.mcp.json` file is written to the Laravel directory rather than the project root. Since Claude Code (and other agents) are typically launched from the project root, the MCP server configuration isn't discovered, and MCP prompts like `/upgrade-laravel-v13` are unavailable.

## Current Behavior

`ClaudeCode::mcpConfigPath()` returns a hardcoded relative path:

```php
public function mcpConfigPath(): string
{
    return '.mcp.json';
}
```

This writes `.mcp.json` to wherever `boost:install` is run from (typically the Laravel app root), with no way to override it.

## Expected Behavior

Support a `mcp_config_path` configuration option in `config/boost.php`, consistent with the existing `guidelines_path` and `skills_path` overrides:

```php
// config/boost.php
'agents' => [
    'claude_code' => [
        'guidelines_path' => '../CLAUDE.md',   // ✅ already supported
        'skills_path' => '.claude/skills',      // ✅ already supported
        'mcp_config_path' => '../.mcp.json',    // ❌ not yet supported
    ],
],
```

## Context

- The `guidelinesPath()` method on the ClaudeCode agent already reads from config: `config('boost.agents.claude_code.guidelines_path', 'CLAUDE.md')`
- The same pattern could be applied to `mcpConfigPath()`
- This would also benefit other agents (Cursor, Copilot, etc.) in monorepo setups

## Environment

- Laravel Boost v2.3.4
- Laravel 12.x
- Monorepo with Laravel app in a subdirectory

johnbacon:
## Summary

Fixes #709

- Adds a `mcp_config_path` config option to all agents, consistent with the existing `guidelines_path` and `skills_path` overrides
- Adds `cwd` to ClaudeCode's MCP server config so `boost:mcp` runs in the correct directory when the config file is placed outside the Laravel root
- Follows the existing pattern from Junie (`config('boost.agents.junie.mcp_path')`) and Codex (`mcpServerConfig()` with `cwd`)

### Usage

```php
// config/boost.php
'agents' => [
    'claude_code' => [
        'guidelines_path' => '../CLAUDE.md',
        'skills_path' => '.claude/skills',
        'mcp_config_path' => '../.mcp.json', // new
    ],
],
```

Without the `cwd` field, placing `.mcp.json` in a parent directory would cause `php artisan boost:mcp` to fail — the command would run from the project root where there is no `artisan` file. The `cwd` field ensures the command executes in the Laravel app directory regardless of where the config file lives.

## Test plan

- [x] Existing test suite passes (724 tests)
- [x] New tests for default/configured `mcpConfigPath()` on ClaudeCode
- [x] New tests for `mcpServerConfig()` including `cwd` field, config override, and empty value filtering

### DIFF
--- a/src/Install/Agents/ClaudeCode.php
+++ b/src/Install/Agents/ClaudeCode.php
@@ -42,6 +42,11 @@ class ClaudeCode extends Agent implements SupportsGuidelines, SupportsMcp, Suppo
         ];
     }
 
+    public function useAbsolutePathForMcp(): bool
+    {
+        return true;
+    }
+
     public function mcpInstallationStrategy(): McpInstallationStrategy
     {
         return McpInstallationStrategy::FILE;


src/Install/Agents/ClaudeCode.php
```diff
@@ -49,7 +49,19 @@ public function mcpInstallationStrategy(): McpInstallationStrategy
...
+    /** {@inheritDoc} */
+    public function mcpServerConfig(string $command, array $args = [], array $env = []): array
+    {
+        return collect([
+            'command' => $command,
+            'args' => $args,
+            'cwd' => config('boost.executable_paths.current_directory', base_path()),
+            'env' => $env,
+        ])->filter(fn ($value): bool => ! in_array($value, [[], null, ''], true))
+            ->toArray();
```
pushpak1300: Any need for this ?

johnbacon: If I'm understanding the question and response properly (apologies if not)...

The issue I ran into was that when configuring an `.mcp.json` to be in a parent directory/project root (like we need to do in our monorepo codebase), Claude Code would try to run the command where there's no `artisan` and the MCP server would fail to start.

The `cwd` field tells the agent where to execute the command, which is the same approach Codex already uses: https://github.com/laravel/boost/blob/1d4f6b1490ff51330e8448eb467db199a35c4002/src/Install/Agents/Codex.php#L74-L84

I suppose because there's now duplication, we could de-duplicate it. But I (and my Claude Code) didn't get into that.

pushpak1300: I checked the Claude Code MCP docs (https://code.claude.com/docs/en/mcp) — `cwd` is not a supported field in `.mcp.json`. The documented format only supports `command`, `args`, and `env`. So this would just be silently ignored by Claude Code.

johnbacon: You're absolutely right. The perils of my trusting AI...

I've updated the implementation to use the same `useAbsolutePathForMcp()` pattern that [Junie does](https://github.com/laravel/boost/blob/61d745a7cc15cac3b1f680ada05f9d732ec70626/src/Install/Agents/Junie.php#L24-L27).

In the ways I can test locally, it now seems satisfactory. Hoping that's the case.

pushpak1300: i am thinking of taking different approach here. this will break the config for existing projects. as your use case is very specific (monorepo) you can edit config after running boost:install manually.

PR #730, opened 2026-03-28
Use configured PHP executable path in ToolExecutor subprocess

### LINKED ISSUE #723: Support Mcp::web() HTTP transport in ToolExecutor
## Feature Request

When using `Mcp::web()` (HTTP transport) with PHP-FPM, `ToolExecutor::buildCommand()` uses `PHP_BINARY` to spawn subprocess tool executions. In a PHP-FPM context, `PHP_BINARY` resolves to `/usr/local/sbin/php-fpm` instead of `/usr/local/bin/php`, causing tool calls to fail.

The `boost.executable_paths.php` config already exists and is used by `GuidelineAssist`, `Agent`, and other install-related code — but `ToolExecutor` doesn't reference it.

## Use Case

Running `Mcp::web('/mcp/boost', Boost::class)` behind nginx + PHP-FPM (e.g. Docker), so that remote clients (Claude Code, Cursor) can connect over HTTP rather than stdio.

## Current Behavior

`ToolExecutor::buildCommand()` hardcodes `PHP_BINARY`:

```php
return [
    PHP_BINARY, // /usr/local/sbin/php-fpm when served via FPM
    base_path('artisan'),
    'boost:execute-tool',
    $toolClass,
    base64_encode(json_encode($arguments)),
];
```

All tool calls fail with php-fpm usage output.

## Suggested Change

```php
return [
    config('boost.executable_paths.php') ?: PHP_BINARY,
    base_path('artisan'),
    // ...
];
```

This would make `ToolExecutor` consistent with the rest of the package and allow `BOOST_PHP_EXECUTABLE_PATH` to fix the issue.

## Environment

- PHP: 8.4, Laravel: 13.1.1, Boost: 2.3.4, MCP: 0.6.3
- Docker (nginx → php-fpm)

pushpak1300:
When running Boost's MCP server via HTTP transport (`Mcp::web()`) behind PHP-FPM, `ToolExecutor` uses `PHP_BINARY` to spawn subprocesses — which resolves to `php-fpm` instead of `php`, breaking all tool calls. The `boost.executable_paths.php` config already exists and is respected elsewhere, but `ToolExecutor` wasn't using it.

Fixes #723

### Approach

- Extracted the command normalization logic (absolute path detection, multi-token command splitting) into a shared `CommandNormalizer` helper so it can be reused
- Updated `ToolExecutor::buildCommand()` to read the configured PHP path with `PHP_BINARY` as fallback
- Added tests for absolute paths with spaces, multi-token wrappers like `herd php`, and the default fallback

### DIFF
--- a/src/Install/Agents/Agent.php
+++ b/src/Install/Agents/Agent.php
@@ -11,6 +11,7 @@ use Laravel\Boost\Install\Enums\McpInstallationStrategy;
 use Laravel\Boost\Install\Enums\Platform;
 use Laravel\Boost\Install\Mcp\FileWriter;
 use Laravel\Boost\Install\Mcp\TomlFileWriter;
+use Laravel\Boost\Support\CommandNormalizer;
 
 abstract class Agent
 {
@@ -270,19 +271,7 @@ abstract class Agent
      */
     protected function normalizeCommand(string $command, array $args = []): array
     {
-        if (str_starts_with($command, '/') || preg_match('#^[a-zA-Z]:[/\\\\]#', $command)) {
-            return [
-                'command' => $command,
-                'args' => $args,
-            ];
-        }
-
-        $parts = str($command)->explode(' ');
-
-        return [
-            'command' => $parts->first(),
-            'args' => $parts->skip(1)->values()->merge($args)->all(),
-        ];
+        return CommandNormalizer::normalize($command, $args);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


dbpolito:
This PR created a regression in case you're using container, my php config is something like `docker compose exec app php`, to run this inside my container... With this new code it will try to run the docker inside the container and fail.

I tried to think in a solution that would fix both cases but didn't think in a very good solution yet... Thought about checking the SAPI, or introduce a new config for php_subprocess...

But TBH, this solution feels a bit off overall because as far as i understand this still require you to set the php config to fix the `Mcp::web()` case.

PR #736, opened 2026-04-01
Fix: Update outdated Livewire paths to v4 standards in SKILL.blade.php

Naimul007A:
**Fixes #735**

This PR updates the Livewire `SKILL.blade.php` file to properly enforce Livewire v4 directory structures. 

Previously, the instructions were telling AI agents to use the legacy v3 `resources/views/livewire/` directory by default. This caused AI assistants to generate files in the wrong locations, breaking the modern v4 `component_namespaces` structure (which separates UI components and full pages).

### Changes Made
* Updated the **Basic Usage** commands to reflect the exact paths created in v4 (SFCs, Pages, and Class-based).
* Overhauled the **Component Format Reference** table to explicitly map out `views/components/`, `views/pages/`, and `views/layouts/`.
* Added a **Critical Directory Rules** section to strictly forbid the AI from defaulting to the old `views/livewire/` directory unless explicitly asked to modify legacy code.

### Testing
- Checked the markdown formatting to ensure it renders correctly.
- (Local test) Fed the updated rules to an AI agent, and it successfully generated a new single-file component in `resources/views/components/` instead of the legacy folder.

### DIFF
--- a/.ai/livewire/4/skill/livewire-development/SKILL.blade.php
+++ b/.ai/livewire/4/skill/livewire-development/SKILL.blade.php
@@ -58,6 +58,8 @@ Before creating a component, check `config/livewire.php` for directory overrides
 | Class-based | `--class` | `{{ $assist->appPath('Livewire/CreatePost.php') }}` | `resources/views/livewire/create-post.blade.php` |
 | View-based | ⚡ prefix (optional, configurable) | — | `resources/views/components/create-post.blade.php` (Blade-only with functional state) |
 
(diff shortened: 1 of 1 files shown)


pushpak1300:
I can see there are some unnecessary formatting changes in the diff can you please keep the changes to fix the issue only ?

Naimul007A:
> I can see there are some unnecessary formatting changes in the diff can you please keep the changes to fix the issue only ?

Thanks for pointing that out! My editor auto-formatted the file on save. I've removed the unrelated formatting changes and updated the PR so it only includes the necessary fix.

Issue #716, opened 2026-03-25
Model list is incomplete


OlinB:
### Boost Version

2.4.0

### Laravel Version

12.55.1

### Agent/IDE Name & Version

copilot CLI 1.0.11, jetbrains AI assistant 253.32098.66 and cursor 2.6.18

### PHP Version

8.4.19

### System Info

Windows 11 with laravel Herd

### Description

The model list returned by `laravel-boost-application-info` is incomplete. 
Only the `User` model is returned

### Steps To Reproduce

ask any AI agent to list models returned by the tool. 

The root cause seems to be the `class_exists(..., false)` in https://github.com/laravel/boost/blob/f4057825ca1c465bf26e8ed927f4734910923c47/src/Install/GuidelineAssist.php#L99 
Patching the `autoload` argument to `true` make the agent see all models.

I haven't opened a PR for this because I want to know if I'm doing something wrong, or if this is intentional

larswolterssoftware:
Having the same issue. When asking Codex to list the models, it looks in the metadata first but that contains an empty list of models.

Output while there are 5 model classes:

```
The Laravel Boost app metadata tool reported no models, so I verified by scanning the app directory for Eloquent model classes.
```

thomasplevy:
I'm also seeing that this affects installation with some conditional guidelines such as 

https://github.com/laravel/boost/blob/39bc6dff1d8803cccefc8a2c1ef5b075c7a4d362/.ai/php/core.blade.php#L6-L8

If I patch this as suggested by @OlinB and run `php artistan boost:install` then the expected strict types guideline is added, otherwise it is not. 

This affects my enums as well too later in the same file

bvdnu:
Hit this too in a long-running `boost:mcp` stdio process (Claude Code + laravel/boost 2.4.4 + Laravel 12.56.0). In our case `application-info` returns a completely empty `"models":[]` — not even the `User` model is returned.

Worth adding to the diagnosis: the static `self::$classes` cache in `GuidelineAssist::discover()` means the bad result from the very first discovery call persists for the entire MCP session. Subsequent MCP calls hit the cache and keep returning the same empty list.

**Reproduction:**

\`\`\`bash
# In a long-running boost:mcp process:
# Tool: mcp__laravel-boost__application-info
# Returns: models: []

# But in a fresh tinker context, the logic works fine:
$ php artisan tinker --execute '
    \$assist = new \Laravel\Boost\Install\GuidelineAssist(
[...]

PR #749, opened 2026-04-06
Fix incomplete model discovery in GuidelineAssist class

iz-ahmad:
fixes issue #716 

`laravel-boost-application-info` was returning an incomplete model list (e.g., only `User`) when asked any AI agent to list the models.

Likely root cause was: `discover()` in `GuidelineAssist.php` used:

```php
class_exists($className, false)
```

here, passing `false` disables autoloading, so any model classes that weren't already loaded in memory - would be silently skipped.
                                                                                                                                       
#### Changes made

* **src/Install/GuidelineAssist.php** — Replaced `class_exists($className, false)` with `class_exists($className)` to allow Composer autoloading, ensuring all _autoloadable_ models are discovered.
* **tests/Unit/Install/GuidelineAssistTest.php** — Added a test verifying that `discover()` detects models that were not previously loaded.
* **tests/Fixtures/app/Models/ExtraModel.php** — Added a simple Eloquent model fixture to validate if the fix is working.

Thanks.

### DIFF
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -96,7 +96,7 @@ class GuidelineAssist
                         continue;
                     }
 
-                    if (class_exists($className, false)) {
+                    if (class_exists($className)) {
                         self::$classes[$className] = $path;
                     }
                 } catch (Throwable) {
--- a/tests/Fixtures/app/Models/ExtraModel.php
+++ b/tests/Fixtures/app/Models/ExtraModel.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Models;
+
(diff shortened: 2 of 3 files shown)


pushpak1300:
This isn’t a reliable way because if there’s an incorrect class throwing warnings and fatal errors the boost won’t start.

pushpak1300:
I’m considering removing this functionality as it adds unnecessary bloat to the context and isn’t helpful.  Furthermore, there are so many edge cases to get it right. Today’s model can intelligently determine which model exists and which doesn’t by searching the codebase.

iz-ahmad:
> Furthermore, there are so many edge cases to get it right. Today’s model can intelligently determine which model exists and which doesn’t by searching the codebase.

Yeah, just saw your PR #754.. I understand, thanks.

Issue #149, opened 2025-08-16
Installation Bug: Please run [./vendor/bin/pest] instead.


pabbec:
### Laravel Package Version

1.0.18

### Laravel Version

12.24.0

### PHP Version

8.2

### System Info

macOS 15.5 (M4 Pro), Herd 1.21.2 (build 51), PHP 8.2. MariaDB.

### Description

Hello,

While attempting to install boost using the boost:install command after selecting all positions, I’m encountering the Pest\Exceptions\InvalidPestCommand error.

The error message is exactly:

$ php artisan boost:install
```

 ██████╗   ██████╗   ██████╗  ███████╗ ████████╗
 ██╔══██╗ ██╔═══██╗ ██╔═══██╗ ██╔════╝ ╚══██╔══╝
 ██████╔╝ ██║   ██║ ██║   ██║ ███████╗    ██║
 ██╔══██╗ ██║   ██║ ██║   ██║ ╚════██║    ██║
 ██████╔╝ ╚██████╔╝ ╚██████╔╝ ███████║    ██║
 ╚═════╝   ╚═════╝   ╚═════╝  ╚══════╝    ╚═╝

  ✦ Laravel Boost :: Install :: We Must Ship ✦ 

 Let's give EcommerceApp a Boost

 ┌ Which code editors do you use to work on EcommerceApp? ┐
 │ Cursor                                                         │
 └────────────────────────────────────────────────────────────────┘

 ┌ Which agents need AI guidelines for EcommerceApp? ───┐
 │ Cursor                                                       │
 └──────────────────────────────────────────────────────────────┘


Pest\Exceptions\InvalidPestCommand 

Please run [./vendor/bin/pest] instead.

```

I'm using:

pestphp/pest                                 3.8.2 
filament/filament                            4.0.1
laradumps/laradumps                   4.4.0
barryvdh/laravel-debugbar           3.16.0 

### Steps To Reproduce

`composer require laravel/boost --dev`
`php artisan boost:install`

AlexisSerneels:
Same error.

```
php artisan boost:install

 ██████╗   ██████╗   ██████╗  ███████╗ ████████╗
 ██╔══██╗ ██╔═══██╗ ██╔═══██╗ ██╔════╝ ╚══██╔══╝
 ██████╔╝ ██║   ██║ ██║   ██║ ███████╗    ██║
 ██╔══██╗ ██║   ██║ ██║   ██║ ╚════██║    ██║
[...]

PR #754, opened 2026-04-06
Remove model discovery and use token-based class parsing

### LINKED ISSUE #716: Model list is incomplete

pushpak1300:
The model list returned by the `ApplicationInfo` MCP tool was broken (#716) — `class_exists($className, false)` only found already-loaded classes, so most models were missing. Enabling autoloading caused crashes (Pest errors, trait redeclarations, fatal errors — see #149, #100). Rather than fixing an endless stream of edge cases, this removes model discovery entirely. Modern AI agents can find models by searching the codebase directly.

Fixes #716

> [!WARNING]
> The `ApplicationInfo` MCP tool no longer returns a `models` key. The `models()` and `shouldEnforceStrictTypes()` methods have been removed from `GuidelineAssist`.

### DIFF
--- a/.ai/php/core.blade.php
+++ b/.ai/php/core.blade.php
@@ -3,9 +3,6 @@
 @php
 /** @var \Laravel\Boost\Install\GuidelineAssist $assist */
 @endphp
-@if($assist->shouldEnforceStrictTypes())
-- Always declare `declare(strict_types=1);` at the top of every `.php` file.
-@endif
 - Always use curly braces for control structures, even for single-line bodies.
 - Use PHP 8 constructor property promotion: `public function __construct(public GitHub $github) { }`. Do not leave empty zero-parameter `__construct()` methods unless the constructor is private.
 - Use explicit return type declarations and type hints for all method parameters: `function isAccessible(User $user, ?string $path = null): bool`
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -52,10 +52,10 @@ class GuidelineAssist
             ->name('/[A-Z].*\.php$/');
 
         foreach ($finder as $file) {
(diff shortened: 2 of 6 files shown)


iz-ahmad:
> The model list returned by the `ApplicationInfo` MCP tool was broken (#716) — `class_exists($className, false)` only found already-loaded classes, so most models were missing. Enabling autoloading caused crashes (Pest errors, trait redeclarations, fatal errors — see #149, #100). Rather than fixing an endless stream of edge cases, this removes model discovery entirely. Modern AI agents can find models by searching the codebase directly.

seems like a fair decision 👍 that feature now doesn't have much necessity left..

benbjurstrom (commented):
Should we update the Resources description as well? Otherwise it looks good.

https://github.com/laravel/boost/blob/646e985abd0b3f1662749e93e2fffe6ecb12dc2f/src/Mcp/Resources/ApplicationInfo.php#L22

PR #766, opened 2026-04-11
Fix model/controller/enum discovery when classes are not pre-loaded

RicardoSemedo:
When Boost runs as a long-running MCP server, `application-info` returns a nearly empty list of models, controllers, and enums for most apps.
On a real Laravel 12 project with 19 models in `app/Models/`, only 1 was discovered — the one Laravel happened to touch during boot.

  ## What's going on

  `GuidelineAssist::discover()` walks `app_path()` with Symfony Finder and checks each file with `class_exists($className, false)`. The second argument disables autoloading, so PHP only "sees" classes that some earlier code path has already loaded.

  During a normal HTTP request that's usually fine — enough code runs to warm up a reasonable set of classes. But when Boost runs as a persistent MCP server, the only classes Laravel loads on boot are the handful reached by the service provider graph. Everything else in `app/Models/` stays dormant on disk and is invisible to `class_exists(..., false)`.

  ## The fix

  One character: flip the flag to `true` so the autoloader can actually resolve the class.

  This is safe because `discover()` already gates each file with `fileHasClassLike()` a few lines above. That helper confirms the file defines a class, interface, trait, or enum before `class_exists` is ever called, so we're only asking the autoloader to resolve files we've already
  verified.

  ## Test

  Added a regression test that writes a throwaway model into a temp directory, registers a runtime PSR-4 mapping with Composer's class loader, points `app_path()` at the temp dir, and asserts `GuidelineAssist::models()` discovers it. The test first confirms the fixture is **not**  already in `get_declared_classes()`, so it genuinely exercises the autoload path.

  I used a runtime temp fixture rather than a static one under `tests/Fixtures/` because `tests/ArchTest.php` runs
  `arch('tests')->expect('Tests')`, which eagerly loads every class under the `Tests\` namespace and would defeat the "not pre-loaded"
  precondition.

  Against the unpatched code the new test fails. With the fix, the full suite is green (737 passed). Pint, Rector, and PHPStan clean.

### DIFF
--- a/src/Install/GuidelineAssist.php
+++ b/src/Install/GuidelineAssist.php
@@ -96,7 +96,7 @@ class GuidelineAssist
                         continue;
                     }
 
-                    if (class_exists($className, false)) {
+                    if (class_exists($className, true)) {
                         self::$classes[$className] = $path;
                     }
                 } catch (Throwable) {
--- a/tests/Feature/Install/GuidelineAssistDiscoverTest.php
+++ b/tests/Feature/Install/GuidelineAssistDiscoverTest.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+use Composer\Autoload\ClassLoader;
+use Laravel\Boost\Install\GuidelineAssist;
+use Laravel\Boost\Install\GuidelineConfig;
+use Laravel\Roster\Enums\NodePackageManager;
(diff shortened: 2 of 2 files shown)


pushpak1300:
we will be deprecating that functionality soon. https://github.com/laravel/boost/pull/754

PR #769, opened 2026-04-13
Testing Best Practices Skill

### LINKED ISSUE #662: Specific instructions/abilities or tool for testing
AI models seem to do fairly well at writing tests but not designing tests - I frequently end up writing a separate instruction set specific to testing and including it as context.  

It would be nice to see boost including include an instruction set specific to designing Laravel tests using accepted best practises and methodology. 

No - this is not Laravel/Boost specific, but it would be nice to not have to find, refactor, write a specific instruction set for every Laravel project I work on. 

- generally AI does not use a coverage report to identify what needs to be tested or is already tested elsewhere (or how to interpret that report)
- AI generates a lot of spurious tests (excessive token use)
- AI generally does not stop on failure, or target specific test classes/methods (more excessive token use) 

something along the lines of: 

```markdown
---
applyTo: "tests/**"
---

# PHPUnit Test Design Methodology

## Test Philosophy

Tests verify behaviour, not implementation. Before writing any test, ask: *if the implementation changed but the outcome stayed the same, would this test still pass?* It should.

Tests must be **FIRST**: Fast, Isolated, Repeatable, Self-validating, Timely. Each test method verifies one thing. A failing test should identify the broken behaviour without ambiguity.

Structure every test using **AAA**: arrange preconditions, act on the SUT, assert the expected outcome — separated by blank lines.

---

## What to Test

For any feature, cover:

- Happy path — expected output, side effects, response status
- Unauthenticated access — redirect or 401
- Unauthorised access — 403, tested against the policy/gate directly, not through a controller that has other logic
- Each required field missing
- Each boundary violation (min, max, type)
- Uniqueness and foreign key constraints — tested behaviourally via HTTP, not by asserting rule strings
- Edge cases — null optionals, empty collections, boundary values

**Behavioural over structural.** Prefer submitting invalid input through the stack and asserting the error, over asserting that a rules array contains a particular string. Structural assertions are only acceptable for methods that cannot be reached via HTTP; document why.

**One assertion purpose per test.** Redundant assertions — checking status 200 before `assertSee`, asserting a value that a prior assertion already implies — add noise without signal. Remove them.

---

## Test Isolation

- Prefer private factory methods over `setUp()` for building the SUT. `setUp()` runs for every test and its allocated objects persist in memory until the suite completes; factory methods use locals that are freed immediately.
- Authorization tests: test policies directly on the policy class, or via a minimal dedicated test route. Do not test authorization by routing through a real controller — a 403 could come from middleware, the policy, or a manual `abort()`, and you cannot tell which.
- **Flag gaps before writing tests.** Before writing tests for any class, audit it for discrepancies between intent and implementation:
  - *Empty stub methods* — a method with no body (or just `//`) has no testable behaviour. Do not write a test for it; instead, add a comment in the test file: `// NOTE: MethodName() is an unimplemented stub — no tests written until it has behaviour.`
  - *Authorization gaps* — if a policy exists for the model but the controller action does not invoke `$this->authorize()`, the effective rule is "any authenticated user", which may be unintentional. Write tests reflecting actual behaviour and note: `// NOTE: FooPolicy::view() exists but action does not call $this->authorize().`
  - *Other gaps* — missing validation on a mutating action, a relationship loaded in one branch but not another, a response that returns inconsistent status codes. Flag anything that looks like it contradicts the declared design.
  
  Report all flagged gaps to the user before or alongside delivering the tests. Do not silently paper over them.
- Use `#[CoversClass(ClassName::class)]` on every test class. This enforces test boundaries, prevents accidental coverage, and scopes `--covers` filtering.
- Use `#[Small]`, `#[Medium]`, `#[Large]` to categorise by scope and execution time.
- Use data providers (`#[DataProvider]` or `#[TestWith]`) for boundary tests rather than repeating near-identical test methods.

---

## Running Tests

These rules extend and take precedence over any running-tests guidance in `copilot-instructions.md`. Prefer targeted execution to keep feedback fast and context compact.

```
# Run a specific file:
php artisan test --compact tests/Feature/Http/Requests/StoreNominationRequestTest.php

# Filter by test class or method name:
php artisan test --compact --filter StoreNominationRequestTest
php artisan test --compact --filter test_label_too_short
```

- Use `--filter` with a test class name or method name for the most immediate targeting.
- Only run the full suite (`php artisan test --compact`) to confirm no regressions after a feature is complete.

---

## Coverage-Guided Testing

Coverage reports show which lines executed — not whether behaviour was verified. A fully green report can hide a suite full of hollow assertions.

Use coverage to find untested **branches** (the `false` side of a condition, an unhandled exception path, a guard clause never triggered), not to reach a line-count target. For each uncovered branch, identify what input or state would cause it to execute, then write a test that creates that state and asserts the resulting outcome.

Coverage at 80% with meaningful assertions is more valuable than 100% with hollow ones.

When using a coverage report as the starting point, generate it first, sort by lowest branch coverage, prioritise critical paths (authorisation, mutation, financial logic), then author tests. Do not author tests bottom-up from the report — use it as a gap-finder, then design tests from the user/behaviour perspective.

---

## Naming

Test method names are executable specifications. They should read as a sentence:

- `test_unauthenticated_request_redirects_to_login`
- `test_non_admin_user_is_forbidden`
- `test_label_too_short_fails_validation`
- `test_valid_payload_creates_record_and_returns_201`

Avoid `test_store`, `test_it_works`, `test_validation`. The name should describe the scenario and expected outcome, not the method being called.

```

nexxai:
Description update by @pushpak1300 

Currently, Boost has a few different testing guidance/skills:

* `pest-testing`
* `enforce-testing`
* `phpunit-guidelines`

They all overlap in different ways, which makes the current testing guidance a bit hit or miss.

We've also had feedback that agents sometimes go too far with tests. They may add tests for tiny or unnecessary changes, over-mock things, test framework behavior or config that doesn't really need testing, or generally write more tests than the change actually requires.

To improve this, we're consolidating the existing guidance into a new `test-best-practices` skill, similar to `laravel-best-practices`.

The goal is to give agents better guidance on what should be tested, how much testing is appropriate, and how to write effective tests for Laravel applications without over-testing.

The skill will also be composed based on the packages installed in the project. For example, it can include specific guidance for:

* Pest
* PHPUnit
* browser testing
* TIA and other testing-related plugins

This should give us one consistent testing philosophy while still adapting the guidance to each application's actual testing stack.

I ended up running new skill on https://github.com/pinkary-project/pinkary.com and found this issue which imo are validate the issues.
<details>
# Test Suite Audit

661 tests, 137 files, ~12.7k lines, reviewed against `.claude/skills/testing-best-practices`.
A report — nothing was changed. Per `rules/review.md` a duplicate **shrinks at the higher layer to the
one case that proves the wiring**, it is not deleted; a pattern the project repeats is a convention,
named once rather than per file.

## Fix first — a real defect ships green today

| # | Where | Gap |
| --- | --- | --- |
| 1 | `app/Models/User.php:97` | `canAccessPanel()` has **no test at all**. Its `hasVerifiedEmail()` branch and second admin (`mrpunyapal@gmail.com`) are uncovered, while 5 Citadel HTTP files repeat the same 2-case matrix by hard-coding an email. → matrix into `UserTest.php`, one refused-role HTTP test per entry point. |
| 7 | `livewire/questions/show.blade.php:58` | `{!! $question->content !!}` renders `ParsableContent` output raw; `Unit/Services/ContentTest.php` feeds it **zero hostile inputs**. `StripProviderParsable` is the only guard and nothing asserts it. → add a case shaped like `BioTest.php:12`. |
| 6b | `Unit/Models/QuestionTest.php:47-71` | `root`/`parent`/`children`/`descendants` add `->where('is_ignored', false)->where('is_reported', false)` (`Question.php:231-266`); the test only checks instance types, so deleting both clauses stays green. → assert an ignored/reported child is excluded. |

## Then — same defect covered twice, shrink the higher layer

| # | Where | Duplicate |
| --- | --- | --- |
| 2 | `Http/Register/CreateTest.php:200-276`, `:333-351` | 32-case + 16-case username datasets run as full HTTP requests; blocked-email repeats `NotBlockedAccountTest`. Split is inverted — `UsernameTest.php:20-32` lumps 6 inputs under one `->throws()` with **no message assertion**. → matrix + messages down to the rule tests, one case per rule stays up. |
| 3 | `Livewire/Links/IndexTest.php:247-267`, `PeopleToFollowTest.php:48-60` | Re-test follow / unfollow / guest-redirect that `Concerns/FollowableTest.php` already owns. |
| 4 | `Http/About/IndexTest.php` + `Http/Verified/IndexTest.php` | Same 4 layout facts (login button, "Your Profile", terms block, version) against one shared partial. |
| 4 | `Http/Bookmarks/IndexTest.php:29-45` | Builds **51 users with links and questions** to re-assert the cut-off `PeopleToFollowRecommendationsTest` owns — likely the suite's most expensive test. |
| 5 | `Http/LoginTest.php:31-42` | Rate-limit test strictly contains the invalid-password test at `:45-56` — 6 bcrypt hashes — and never asserts `assertGuest()` on the 429. |

## Then — suite-wide hygiene

| # | Finding |
| --- | --- |
| 8 | `tests/Pest.php`: no `Http::preventStrayRequests()` (0 suite-wide — a missing fake reaches the network), no `Sleep::fake()`, no `Exceptions::fake()`; `RefreshDatabase` over `LazilyRefreshDatabase`; global `Storage::fake()` in `beforeEach()`; two dead stubs (`something()` at `:50`, `toBeOne()` at `:37`). |
| 10 | 52 × `assertStatus(200\|302\|403\|404\|429)` across 20 files instead of `assertOk()`/`assertForbidden()`/`assertNotFound()`/`assertTooManyRequests()`. Plus `Citadel/Users/IndexTest.php:24` uses `assertDatabaseHas` where `assertModelExists()` applies. |
| 11 | 45 names give no result — 23 are literally `'guest'`/`'auth'`, the rest bare nouns (`'relations'`, `'link'`, `'mention'`). 11 files mix `it()` and `test()`; worst is `Livewire/Questions/CreateTest.php` (1 vs 50). |
| 12 | Factory records built in `beforeEach()`: `Livewire/Questions/EditTest.php:12`, `CreateTest.php:20-21`, `Http/Profile/ShowTest.php:11`. |
| 14 | `Http/Register/CreateTest.php:302-325`: AAA comments instead of blank lines, a `// or specific route if expected` leftover, a destination-less `assertRedirect()`, and `expect($user)->not->toBeNull()` after `assertDatabaseHas` proved it. |

## Your call — repeated conventions, not fixes

| # | Pattern | Cost |
| --- | --- | --- |
| 6a | `test('relations')` in 6 model files asserting `toBeInstanceOf` on unconstrained relations | Tests Eloquent, not this project. |
| 13 | 8 × `toMatchSnapshot()`, 4 of them the whole of `ContentTest.php` | States no contract; an XSS regression surfaces as a diff a reviewer may bless. |
| 9 | 9 × `expect(true)->toBeTrue()` across all 7 `Unit/Rules/*` files | The real check is `$this->fail()` in the closure; reads as if nothing is asserted. |

## Checked and correct — no action

`arch()` tests (judged by the convention they protect, not behavioral value) · the 9 `test('to array')`
model tests (`rules/isolation.md` endorses asserting exact `toArray()` keys; each uses `toHaveCount()`) ·
`BCRYPT_ROUNDS=4` in `phpunit.xml` · bio XSS escaping at `Unit/Services/BioTest.php:12`.

</details>

> [!NOTE]
> Once merged we need to delete this [skill](https://github.com/pestphp/pest/tree/5.x/resources/boost/skills/pest-testing)


jackwh:
I think it should be `Http::preventStrayRequests()` (which prevents them), rather than `Http::preventingStrayRequests()` (which returns a bool) 👀 FYI!

PR #770, opened 2026-04-14
fix: call test binary directly in determineTestEnforcement()

### LINKED ISSUE #512: InstallCommand::determineTestEnforcement() times out when listing tests
### Boost Version

2.0.5

### Laravel Version

12.44.0

### Agent/IDE Name & Version

_No response_

### PHP Version

8.4

### System Info

macOS Sequoia Apple M3 Max, Laravel Herd

### Description

We have roughly 16,000 PHPUnit tests in our project, and we've run into this issue each time we've tried to add Boost to the project (first around 3 months ago, and most recently today).

When running `boost:install`, the install determines if tests are present in the project by calling `php artisan test --list-tests` using the Symfony Process component. It has the default timeout of 60 seconds applied, and it hits this timeout each time we try.

If I run the command manually in the terminal it completes in 1 second. Increasing the Process timeout has no effect, it seems as if it's hanging indefinitely.

If I change `determineTestEnforcement()` to instead call `phpunit` directly, it succeeds:
```php
$process = new Process([PHP_BINARY, 'vendor/phpunit/phpunit/phpunit', '--list-tests'], base_path());
```

There are couple of options to address this:
- Detect Pest or PHPUnit and call the binary directly. Boost already detects whenever the Pest or PHPUnit packages are installed.
- Add a prompt to ask the user manually if they want to include testing guidelines.

### Steps To Reproduce

Honestly I'm not sure; there could be something specific to our codebase that is modifying the output of `php artisan test --list-tests` in a way that affects the process component.

yousefkadah:
Fixes #512

## Problem

`determineTestEnforcement()` runs `php artisan test --list-tests` via Symfony Process, which hangs indefinitely on projects with large test suites (16,000+ tests). The issue is specific to routing through `artisan` — calling the test binary directly works fine.

## Solution

- **Call the test binary directly** — detect whether Pest or PHPUnit is installed and invoke the binary (`vendor/bin/pest` or `vendor/bin/phpunit`) instead of going through `php artisan test`
- **Add a 30-second explicit timeout** as a safety net (Process default is 60s, but explicit is clearer)
- **Gracefully return `false`** if the process fails or times out, instead of proceeding with empty output

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -159,8 +159,17 @@ class InstallCommand extends Command
             return false;
         }
 
-        $process = new Process([PHP_BINARY, 'artisan', 'test', '--list-tests'], base_path());
+        $binary = file_exists(base_path('vendor/bin/pest'))
+            ? base_path('vendor/bin/pest')
+            : base_path('vendor/bin/phpunit');
+
+        $process = new Process([PHP_BINARY, $binary, '--list-tests'], base_path());
+        $process->setTimeout(30);
         $process->run();
+
+        if (! $process->isSuccessful()) {
+            return false;
+        }
 
         return Str::of($process->getOutput())
             ->trim()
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for your pull request!

PR #774, opened 2026-04-16
Move deployment guideline

pushpak1300:
This PR moves the guidelines from the core guidelines to deployment guidelines. 

If anyone want to override they can override it via. `.ai/deployments/core.blade.php` and add there own instructions.  

You can also ignore the guidelines by adding this config in `boost.php`

```
'guidelines' => [
   'exclude' => [
       'deployments'
    ]
]
```


rforced:
@pushpak1300 is there a full list of exclude-able guidelines somewhere? The docs on laravel website are still very bare-bones.

PR #776, opened 2026-04-17
Add Laravel Cloud integration to install command

pushpak1300:
Extends the Boost installer with Laravel Cloud support so projects can automatically pull down the `deploying-laravel-cloud skill` alongside their other integrations.

Screenshot 2026-04-17 at 10 41 55 PM


Frostist:
This would be a nice addition!

PR #779, opened 2026-04-20
The path to the Junie guidelines file has been corrected

andrey-helldar:
Junie can't read the `AGENTS.md` file from the project root, so the rules defined there are ignored.

Junie expects the file `.junie/AGENTS.md`.

Since the path key is located deep in the configuration (`boost.agents.junie.guidelines_path`), you can't simply specify the path - Laravel will overwrite other configurations due to the way the `array_merge` function works.

Junie also doesn't allow you to set global project settings, so you'll have to manually change the file path in 100+ projects.

Instead, I suggest changing the default path in Laravel Boost.

Junie version 261.1218.175

image

### DIFF
--- a/src/Install/Agents/Junie.php
+++ b/src/Install/Agents/Junie.php
@@ -73,7 +73,7 @@ class Junie extends Agent implements SupportsGuidelines, SupportsMcp, SupportsSk
 
     public function guidelinesPath(): string
     {
-        return config('boost.agents.junie.guidelines_path', 'AGENTS.md');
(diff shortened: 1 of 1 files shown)


pushpak1300:
Thanks for your pull request!

Unfortunately, I'm going to delay merging this code for now as it introduces a breaking change to the package.

https://junie.jetbrains.com/docs/guidelines-and-memory.html#how-junie-cli-discovers-guidelines

PR #798, opened 2026-05-08
Add optional Laravel Boost AI artifact ignores during `boost:install`

SagarNaliyapara:
## Summary

Adds a small DX improvement to `php artisan boost:install` by offering to append recommended AI artifact entries to `.gitignore`.

If confirmed, Boost adds this grouped section:

```gitignore
# Laravel Boost
.ai/generated
.ai/cache
.claude/
.cursor/rules/generated
```

The update is safe and idempotent:

- creates `.gitignore` if it does not exist
- preserves existing file contents and ordering
- appends only missing entries
- avoids duplicate section headers
- skips the prompt in non-interactive mode

## Implementation

- keeps the logic inline in `InstallCommand.php`
- follows the command’s existing confirmation, output, and file-handling patterns
- does not introduce new services, config, or abstractions

## Tests

Added focused coverage for:

- appending entries to an existing `.gitignore`
- creating `.gitignore` when missing
- avoiding duplicates when the Boost section already exists

## Note

This PR is intended to provide a convenient starting point for AI-related `.gitignore` entries during `boost:install`, not to define a perfect universal ignore list.

I’m not fully sure the current list for .gitignore is ideal for every workflow or at all.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -88,6 +88,7 @@ class InstallCommand extends Command
         $this->discoverEnvironment();
         $this->collectInstallationPreferences();
         $this->performInstallation();
+        $this->updateGitignore();
         $this->outro();
 
         return self::SUCCESS;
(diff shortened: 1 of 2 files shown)


Riari:
I shared this feedback elsewhere but I'll include it here too: I think, if anything, the command should just print the suggested .gitignore entries in interactive sessions. Anything beyond that is over-engineered IMO, especially as the docs already cover this.

pushpak1300:
Yes I agree adding entries directly to `.gitignore` is not a good option. we can just display this after install command.

Patabugen:
A few thoughts based on making an (internal) Skill which sets up a project for agentic development which does this too.

The exact list is probably dev-specific which might add some nuance to the PR, but I think it's good in principle.

Our company standard is to use `.ai/` for things which are committed to the repo (although admittedly we don't have skills yet but I imagine those can go in `.ai/skills`.

All the other vendor-specific directories get gitignored.

It certainly doesn't make sense to add `.claude` but none of the other vendors - that's a highly personal (if popular) choice.

In principle I do agree that Boost should helps manage gitignoring the files it generates (and recommends are gitignored) and adding `boost.json` seems like an obvious choice.

Here's my list of what we add to .gitignore:

```
.agents
.mcp.json
boost.json
opencode.json
CLAUDE.md
.gemini
.cursor
.codex
.junie
GEMINI.md
AGENTS.md
.claude/
.context/
polyscope.json
.ai/guidelines/*.local.md
.github/skills
```

Issue #728, opened 2026-03-27
Default AI guidelines suggest a security vulnerability


ejungcl:
Copilot flagged an issue in the CLAUDE.md and suggested this change.
```
- To check environment variables, read the `.env` file directly.
+ To check environment-dependent settings, prefer `php artisan config:show` and inspecting the `config/*.php` files. Never open, paste, or share the contents of the `.env` file in chat, as it may contain secrets.
```

pushpak1300:
This guideline is intentional — it steers the model on where to find environment variables. We previously had a dedicated environment variable tool in Boost which was removed, so this is the current approach.

Agents already have file system access and can read `.env` regardless of what the guideline says, so there's no security escalation here. If your agent supports config-level file access restrictions, you can block `.env` reads there.

PR #801, opened 2026-05-11
Remove redundant guidance to read .env directly

loadinglucian:
Thanks to the recent core-guideline cleanup work in #722 and #787 — the file reads cleanly. I'd like to propose removing one more redundant line: the instruction to read `.env` directly to check environment variables.

The line immediately above already documents `config:show` and direct reads of `config/*.php`, which cover the same use case. After tracing this through, the `.env` line looks redundant for legitimate lookups while pointing the model at a file Anthropic's [secure-deployment guide](https://code.claude.com/docs/en/agent-sdk/secure-deployment) and [settings reference](https://code.claude.com/docs/en/settings) treat as a sensitive denylist target alongside `~/.aws/credentials` and `*.pem`.

I saw the prior discussion in #728 and agree with the response there: guideline text can't enforce access restrictions, that belongs in the agent's permission layer. This PR isn't trying to revisit that. It's a removal in the same spirit as  #787, where a section was deleted once an adjacent mechanism made it redundant.

Behavior for users workflow is unchanged: agents can still read `.env` files.


pushpak1300:
Fair enough. can be removed. Thanks.

PR #629, opened 2026-03-05
Remove Artisan wrapper MCP tools and update guidelines to use CLI directly

pushpak1300:
Six MCP tools that are thin wrappers around Artisan commands or simple config/env reads are removed. Modern AI agents can invoke these commands directly via shell, so the MCP tool overhead is unnecessary. Guidelines are updated to teach agents the direct CLI equivalents.

### Approach

- Removed 6 tools: ListArtisanCommands, ListRoutes, Tinker, GetConfig, ListAvailableEnvVars, ListAvailableConfigKeys
- Updated guidelines to point agents to `php artisan list`, `--help`, `tinker --execute`, `config:show`, `route:list`, and direct `.env` file reading
- Updated integration tests to use DatabaseConnections/DatabaseQuery as test subjects
- Kept 9 high-value tools that provide unique capabilities not available via CLI

> [!WARNING]
> Breaking change: The `list-artisan-commands`, `list-routes`, `tinker`, `get-config`, `list-available-env-vars`, and `list-available-config-keys` MCP tools are no longer available. Agents should use the equivalent Artisan commands directly.

PR #807, opened 2026-05-15
Add Tinker MCP tool (opt-in via config)

### LINKED ISSUE #667: Missing Tinker MCP Tool
### Boost Version

2.3.1

### Laravel Version

12

### Agent/IDE Name & Version

Cursor/Codex

### PHP Version

8.5

### System Info

Ubuntu 22.04

### Description

There used to be a tinker tool that AI could use. It is now replaced by guidelines to be:
`php artisan tinker --execute "your code here"`

The issue is AI very reliably used to use the tinker tool. Now with the new way, it often fails 3-4x with syntax issue before finally getting it right. This has been the case since the intial boost release. Is it possible to have the old tool again as opt-in atleast 

### Steps To Reproduce

Ask the AI to do something slighty complex like queries or big chunk of code in tinker with tool and direct and it will always take 2-3 tries via cli compared to tool working right away

pushpak1300:
Currently the `tinker` MCP tool is gone (removed in #629), so agents have to use `php artisan tinker --execute '...'` via shell. That round-trips through shell escaping and bites pretty consistently. As #667 points out, agents tend to fail 2-3 times on quoting before they get a snippet right.

This PR brings the tool back, but opt-in so it stays off by default. Flip it on with:

\`\`\`php
// config/boost.php
'tinker_tool_enabled' => true,
\`\`\`

 `core.blade.php` swaps the guidance to prefer the MCP tool when the flag is on, and keeps the existing `--execute` instructions otherwise.
 Closes #667.

### DIFF
--- a/.ai/boost/core.blade.php
+++ b/.ai/boost/core.blade.php
@@ -29,5 +29,9 @@
 
 ## Tinker
 - Execute PHP in app context for debugging and testing code. Do not create models without user approval, prefer tests with factories instead. Prefer existing Artisan commands over custom tinker code.
+@if($assist->hasMcpEnabled() && config('boost.tinker_tool_enabled', false))
+- Use the `tinker` MCP tool to execute PHP code instead of the CLI. It avoids shell escaping issues and runs the snippet in the Laravel application context.
+@else
 - Always use single quotes to prevent shell expansion: `{{ $assist->artisanCommand("tinker --execute 'Your::code();'") }}`
   - Double quotes for PHP strings inside: `{{ $assist->artisanCommand("tinker --execute 'User::where(\"active\", true)->count();'") }}`
+@endif
--- a/src/Mcp/Boost.php
+++ b/src/Mcp/Boost.php
@@ -18,6 +18,7 @@ use Laravel\Boost\Mcp\Tools\GetAbsoluteUrl;
 use Laravel\Boost\Mcp\Tools\LastError;
 use Laravel\Boost\Mcp\Tools\ReadLogEntries;
 use Laravel\Boost\Mcp\Tools\SearchDocs;
+use Laravel\Boost\Mcp\Tools\Tinker;
 use Laravel\Mcp\Server;
 use Laravel\Mcp\Server\Prompt;
 use Laravel\Mcp\Server\Resource;
@@ -91,6 +92,7 @@ class Boost extends Server
             LastError::class,
             ReadLogEntries::class,
             SearchDocs::class,
+            Tinker::class,
         ], 'tools');
     }
 
--- a/src/Mcp/Tools/Tinker.php
+++ b/src/Mcp/Tools/Tinker.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Boost\Mcp\Tools;
+
+use Illuminate\Contracts\JsonSchema\JsonSchema;
+use Illuminate\JsonSchema\Types\Type;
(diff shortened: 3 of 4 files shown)


TonsiTT:
Thanks. Going to switch to this branch.
To show how much it sucked, here is a example of codex trying to use Tinker Tool. It tried 11 times and failed each time and it literally ended up encoding the whole code block to base64 and running it via decode on Tinker as it couldn't figure out how to use the cli version with normal code block due to formatting issue

> Ran 11 commands, used Laravel Boost
> Ran php artisan tinker
> The interactive Tinker paste split the closure at a ternary line, so that attempt didn’t run the validation loop. I’m switching to a single --execute eval(base64_decode(...)) payload so the same app-context code runs as one PHP block.
> Ran 2 commands
> Ran base64 -w0 <<'PHP' $categor....
> Ran php artisan tinker --execute='eval(base64_decode("JGNhdGVnb3J5S.......=="));

pushpak1300:
which model are you using ? @TonsiTT

rforced:
@pushpak1300 where does that configuration go in boost.php?

`'tinker_tool_enabled' => true,`

Is it nested under another key?

pushpak1300:
nope @rforced

rforced:
@pushpak1300 can you comment on why the setting isn't nested under the MCP settings? Right now we have this. Having this setting at top level is starting to make boost's configuration feel very inconsistent

```
    'mcp' => [
        'tools' => [
            'exclude' => [
                GetAbsoluteUrl::class,
            ],
        ],
    ],
```

pushpak1300:
both can work. this is experimental feature. so i haven't documented anything.

rforced:
I understand, thank you

TonsiTT:
> which model are you using ? @TonsiTT

codex on xhigh. Thanks for the update, its back to one-shotting stuff on tinker tool instead of the hoops with cli tool

yusufkandemir:
@pushpak1300

I was facing many issues with the CLI. This tool works nicely, except for some problems that I will document at the bottom. I then took a deeper dive into understanding the differences between CLI behavior and MCP, and figured out that the bigger problem was in my environment. I was using [Spin](https://serversideup.net/open-source/spin/) (wraps Docker Compose similar to Laravel Sail), and it had a bug in argument passthrough, which caused 3-5 tries and falling back to writing a file and executing it. So, it can be that some people who were reporting problems faced similar issues, and wanted to leave here if it helps anybody. In a normal setup, it could still be common to face issues, but maybe in an acceptable amount. I think MCP provides a guaranteed solution to potential issues. Anyways, the things that can be improved are below:

**MCP tool limitations over CLI**:
- `dump([...]);` - tool returns `Invalid JSON output from tool process: Syntax error`. No output, no clue.
- `dd([...]);` - tool returns `Process tool execution failed: <the dump body>`. Output is there, but it is framed as a failure, so the agent treats it as an error.
- Possibly not a big deal, but stderr handling is not proper. If the tinker code the agent wrote or any other executed code in the stack writes to stderr (e.g. `fwrite(STDERR, "...")`):
  - if only written to stderr -> no output at all
  - if both stdout and stderr -> `Invalid JSON output from tool process: Syntax error`

**Common behavior**:
- `$result;` (last line a bare expression) or `return $result;` -> no output. This works fine in interactive mode, but not in `--execute` or the MCP tool. The agent has to remember to use `echo`/`print_r`/etc. explicitly. Sometimes they rely on return or bare expression, and re-run the command/tool again with an explicit output mechanism to get the output.

PR #810, opened 2026-05-18
Add configurable TTL for LastError cache

rpsohag:
## Summary
- Add `last_error_cache_ttl` configuration option
- Default value: 86400 seconds (24 hours)
- Configurable via `BOOST_LAST_ERROR_CACHE_TTL` environment variable
- Set to `null` to cache forever (original behavior)

## Motivation
The LastError tool currently uses `Cache::forever()` which means cached
errors never expire. For long-running applications, this may be undesirable
as old errors can linger indefinitely.

## Type of Change
- [ ] Bug fix
- [x] Feature enhancement
- [ ] Breaking change
- [ ] Documentation update

## Testing
- Added test for default TTL configuration value (86400 seconds)
- Added test for custom TTL configuration
- All existing tests pass (12 tests, 41 assertions)

## Checklist
- [x] Code follows Laravel Boost coding standards (Pint ✓)
- [x] Code passes PHPStan analysis (Level 5 ✓)
- [x] Tests added/updated
- [x] No breaking changes

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -49,4 +49,17 @@ return [
         'current_directory' => env('BOOST_CURRENT_DIRECTORY_EXECUTABLE_PATH'),
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Boost Last Error Cache TTL
+    |--------------------------------------------------------------------------
+    |
+    | This option controls how long (in seconds) the last error is cached.
+    | The cache is used to quickly retrieve the most recent error without
+    | reading from the log file. Set to null to cache forever (default behavior).
+    |
+    */
+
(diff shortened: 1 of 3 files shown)


pushpak1300:
What exactly the problem we are solving ? I am assuming `LastError` should be cached indefinitely.

rpsohag:
Thanks for the feedback! You're right — I didn't fully consider the semantics here.

The "last error" conceptually means "the most recent error in this application," 
which should persist indefinitely until a new error replaces it. Having it expire 
doesn't really align with that purpose, and the cache gets refreshed anyway whenever 
a new error occurs.

I added the TTL thinking old errors might become stale, but you're correct that 
the current behavior (Cache::forever) is actually more appropriate for this use case.

Would you like me to close this PR? I learned a lot from the contribution process 
regardless, and I appreciate you taking the time to review it.

PR #811, opened 2026-05-18
Filter roster to direct requirements only

jlswanson28694:
Boost uses Laravel Roster to determine which packages are installed in the current environment, and uses the results to add guidelines, offer MCP tools, and search documentation. However, it includes indirect packages in the results. I don't have Pint installed in my environment, but Pint keeps showing up in my guidelines because other Laravel packages require it. This PR filters out indirect packages when checking the current Laravel environment.

### DIFF
--- a/.ai/foundation.blade.php
+++ b/.ai/foundation.blade.php
@@ -9,7 +9,7 @@ The Laravel Boost guidelines are specifically curated by Laravel maintainers for
 This application is a Laravel application and its main Laravel ecosystems package & versions are below. You are an expert with them all. Ensure you abide by these specific packages & versions.
 
 - php - {{ PHP_MAJOR_VERSION }}.{{ PHP_MINOR_VERSION }}
-@foreach (app(\Laravel\Roster\Roster::class)->packages()->unique(fn ($package) => $package->rawName()) as $package)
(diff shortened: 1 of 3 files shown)


pushpak1300:
Frankly, the current approach is correct.  If someone uses filament without explicitly adding a livewire we expect it to add livewire guidleines.  Currently we can only identify prod direct dependency because pint will be dev deps in most of the project.

jlswanson28694:
I see what you're saying about the Filament-Livewire relationship, I agree the Livewire guidelines should be added. Perhaps Boost should only filter out specific indirect dependencies?

I didn't catch that indirect dev dependencies don't show up in Laravel Roster. Looks like the reason Pint guidelines keep showing up for me is because Laravel Ranger is including Pint as a prod dependency. Switching that to `require-dev` could be the best way to solve that issue at least. However, as another example, Laravel Prompts is a prod dependency of both Boost and the Laravel Framework, which obviously can't be switched. I don't want Prompts showing up in my guidelines either.

jlswanson28694:
@pushpak1300 I completely overlooked that this problem was already addressed with `DiscoverPackagePaths::mustBeDirect`. I added Pint and Prompts to the array, and it is now used to filter out packages from the foundational context.

PR #813, opened 2026-05-19
Add optional Streamable HTTP MCP transport (disabled by default)

ramhaidar:
## Summary

Adds an optional **Streamable HTTP** transport for Boost's MCP server, alongside the existing stdio transport. Stdio remains the default and recommended transport — nothing changes for current users.

The HTTP transport is **disabled by default** and must be opted into via config/env.

## Motivation

Some MCP clients do not keep stdio child processes alive reliably. After periods of inactivity they kill or lose the `boost:mcp` process and Boost appears disconnected until the client is restarted.

For those clients (and for setups where it is more convenient to point an MCP client at a long-running web server like Herd, Valet, Sail, nginx + PHP-FPM, or `php artisan serve`), Boost can now expose the same MCP server over HTTP using `laravel/mcp`'s built-in `Mcp::web(...)` transport.

### Transport registration

- **`config/boost.php`** — New `mcp.web` block:
  ```php
  'mcp' => [
      'web' => [
          'enabled'    => env('BOOST_MCP_WEB_ENABLED', false),
          'path'       => env('BOOST_MCP_WEB_PATH', '/_boost/mcp'),
          'middleware' => [],
      ],
  ],
  ```
- **`src/BoostServiceProvider.php`** — `boot()` still calls `Mcp::local('laravel-boost', Boost::class)` exactly as before. A new `registerWebMcp()` method conditionally registers the HTTP route via `Mcp::web($path, Boost::class)` only when `boost.mcp.web.enabled` is `true`, applying configured middleware when non-empty.

### Tool execution under HTTP

- **`src/Mcp/ToolExecutor.php`** — When Boost is served over a non-CLI SAPI (PHP-FPM, CGI, `php artisan serve`), `ToolExecutor` now resolves and runs the tool **inline** instead of spawning `PHP_BINARY artisan boost:execute-tool ...`. Under HTTP, `PHP_BINARY` does not point at a CLI php that can register the `boost:` artisan namespace, so the previous subprocess approach failed with `There are no commands defined in the "boost" namespace.` and CGI headers polluting stdout.
  - New `shouldRunInSubprocess()` returns `true` only for `cli` / `phpdbg` SAPI.
  - New `executeInline($toolClass, $arguments)` resolves the tool from the container, invokes `$tool->handle($request)` directly, wraps `Throwable` as `Response::error(...)`, and `ob_*`-buffers stray stdout.
  - Subprocess execution path (used by stdio) is unchanged. Each HTTP request already runs in its own PHP-FPM worker, so subprocess-level isolation is not needed there.

### Tests

- **`tests/Feature/Mcp/WebMcpTest.php`** — covers:
  - default-disabled (no route registered)
  - explicit-disabled
  - enabled registers the route
  - custom path is respected
  - middleware is applied
  - boost-disabled short-circuits HTTP registration
  - stdio + HTTP coexist when HTTP is enabled
  - stdio remains registered when HTTP is disabled
- **`tests/Feature/Mcp/ToolExecutorInlineTest.php`** — covers:
  - non-CLI path resolves a real tool without invoking `buildCommand` / `executeInSubprocess`
  - non-CLI path still rejects unregistered tools
  - inline path wraps thrown exceptions as error responses
  - `shouldRunInSubprocess()` matches the current SAPI

### Docs

- **`README.md`** — new "Optional Streamable HTTP MCP Transport" section with enablement steps, configuration shape, sample MCP client URL, middleware example, and an explicit security warning.

## Usage

Stdio (unchanged, default):
```
php artisan boost:mcp
```

HTTP (opt-in):
```env
BOOST_MCP_WEB_ENABLED=true
BOOST_MCP_WEB_PATH=/_boost/mcp
```
Serve the app normally and point the MCP client at e.g. `http://my-app.test/_boost/mcp`.

Optional middleware:
```php
'middleware' => ['auth:sanctum'],
```

## Security

Boost exposes development tooling that can inspect the application, database, logs, routes, configuration, and other sensitive local-dev information. The HTTP endpoint is therefore **disabled by default** and the README clearly warns it should only be enabled in trusted local environments unless protected with authentication and network-level controls. No auth middleware is prescribed by default since the appropriate choice depends on the host application.

## Backward compatibility

- No change to default behavior.
- `Mcp::local('laravel-boost', Boost::class)` registration is untouched.
- `boost:mcp`, `mcp:start laravel-boost`, and `boost:execute-tool` all behave identically under CLI/stdio.
- HTTP route is only registered when explicitly opted in via config.
- All previously existing `ToolExecutor` tests (subprocess execution, isolation, autoload reload, timeout clamping, `buildCommand` behavior) still pass.

## Tests

- `pest tests/Feature/Mcp/ToolExecutorInlineTest.php tests/Feature/Mcp/ToolExecutorTest.php tests/Feature/Mcp/WebMcpTest.php tests/Feature/BoostServiceProviderTest.php` → **32 passed (66 assertions)**.
- Targeted: `WebMcpTest` 9/9, `ToolExecutorInlineTest` 4/4, `BoostServiceProviderTest` 9/9, `ToolExecutorTest` 10/10.
- Full suite (`pest`): 747 passed. The 5 unrelated failures on Windows (CRLF line-ending diff and `symlink(): Permission denied`) reproduce on `main` without these changes.

## Lint / Static analysis

- `phpstan analyse src` → no errors.
- `pint --format agent` applied to changed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,57 @@ Laravel Boost accelerates AI-assisted development by providing the essential con
 
 Documentation for Laravel Boost can be found on the [Laravel website](https://laravel.com/docs/boost).
 
+## Optional Streamable HTTP MCP Transport
+
+By default, Laravel Boost exposes its MCP server over **stdio** via the `php artisan boost:mcp` command. This is the recommended transport for most local setups and remains unchanged.
+
+Some MCP clients do not keep stdio servers alive reliably and may appear to "disconnect" Boost after periods of inactivity. For those clients, Boost can also expose the same MCP server over HTTP (Streamable HTTP) using `laravel/mcp`'s web transport.
+
+The HTTP transport is **disabled by default**.
+
+### Enabling
+
+Publish the Boost config (if you have not already) and set the env vars:
+
+```env
+BOOST_MCP_WEB_ENABLED=true
+BOOST_MCP_WEB_PATH=/_boost/mcp
+```
+
+Then serve the application normally with your usual local web stack (Herd, Valet, Sail, `nginx`/`php-fpm`, or `php artisan serve`). The MCP client should be pointed at the application URL plus the configured path, for example:
+
+```
+http://my-app.test/_boost/mcp
+```
+
+### Configuration
+
+The full config block (in `config/boost.php`) looks like:
+
+```php
+'mcp' => [
+    'web' => [
+        'enabled'    => env('BOOST_MCP_WEB_ENABLED', false),
+        'path'       => env('BOOST_MCP_WEB_PATH', '/_boost/mcp'),
+        'middleware' => [],
+    ],
+],
+```
+
+You may attach project-appropriate middleware to the HTTP MCP route, for example:
+
+```php
+'middleware' => ['auth:sanctum'],
+```
+
+The stdio transport is unaffected by this setting and continues to work as before.
+
+### Security
+
+> **Warning** Laravel Boost exposes development tooling that can inspect your application, database, logs, routes, configuration, and other sensitive local development information. Do **not** expose the HTTP MCP endpoint on a public network without authentication and network-level protections.
+
+The HTTP transport is intended for trusted local development environments. If you must expose it beyond your local machine, protect it with appropriate middleware (such as `auth:sanctum`) and network-level controls. Do not enable it in production.
+
 ## Contributing
 
 Thank you for considering contributing to Boost! The contribution guide can be found in the [Laravel documentation](https://laravel.com/docs/contributions).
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks for your pull request!

ramhaidar:
Thanks for reviewing @pushpak1300 

Understood on keeping Boost’s core surface area small. I’ll move this out of Boost core and explore shipping it as a separate package that registers the Boost MCP server through Laravel MCP’s web transport in an opt-in way.

One question before I go that route: is relying on `Laravel\Boost\Boost` as the MCP server class acceptable from an external package, or would you prefer Boost expose a small public extension point for registering alternate MCP transports?

Thanks again.

ramhaidar:
External Package: https://github.com/ramhaidar/laravel-boost-streamable-http

PR #839, opened 2026-06-04
fix: guard non-interactive mode in selectThirdPartyPackages and selectIntegrations

### LINKED ISSUE #800: Two bugs in v2.4.6: --no-interaction validation + $N literal mangling
## Summary

Hit two distinct regressions in `laravel/boost@v2.4.6`. Filing combined per project convention; happy to split if preferred.

**Environment:**
- `laravel/boost`: v2.4.6
- `laravel/framework`: v12.58.0
- PHP: 8.5
- Host: WSL2 / Docker Compose v2 (boost commands run inside per-app PHP-FPM containers)

---

## Bug 1: `boost:install --no-interaction` fails validation on `selectAgents()` multiselect

**Reproduction:**

```bash
docker compose exec <app> php artisan boost:install --no-interaction
```

**Expected:** non-interactive install completes using sensible defaults (e.g. `claude_code` only) or skips agent selection.

**Actual:** the `selectAgents()` multiselect prompt still renders interactively, validation fires `Required.`, and the command exits non-zero. The `--no-interaction` flag is effectively ignored at this prompt.

**Workaround:** hand-write `boost.json` with the desired `"agents"` array. Works but defeats the purpose of automated provisioning.

**Suspected cause:** `selectAgents()` multiselect doesn't bind a non-interactive default through `Laravel\Prompts`.

---

## Bug 2: `boost:update` mangles literal `$N` strings when regenerating per-app `CLAUDE.md` from `.ai/guidelines/*.md` Blade-rendered source

**Reproduction setup:**

In `<app>/.ai/guidelines/project.md`, include a literal dollar amount:

```markdown
- Apple Developer Program renewal ($99/yr) before the anniversary
```

**Run:**

```bash
docker compose exec <app> php artisan boost:update
```

**Expected:** `<app>/CLAUDE.md` regenerates with `$99/yr` intact.

**Actual:** rendered output strips `$99`, leaving `(/yr)`:

```diff
- Apple Developer Program renewal ($99/yr) before the anniversary
+ Apple Developer Program renewal ((/yr) before the anniversary
```

**Suspected cause:** Guidelines render through Blade (`RendersBladeGuidelines` trait). Blade's PHP variable interpolation eats `$99` as `$9` (variable → empty) followed by `9`. The visible artifact is `$N` digits silently disappearing in regenerated CLAUDE.md.

This silently affects any guideline that mentions `$N` literals — dollar amounts, regex patterns, AWS-CLI `$1` capture groups, jQuery references, etc.

**Workaround:** Blade-escape every `$` as `@$` or `\$` in source guidelines. Brittle and easy to miss; fails silently on next regen.

**Suggested fix (pick one):**
- Render guidelines as plain Markdown (no Blade) — they don't appear to need Blade features
- Wrap user guideline content in `@verbatim` blocks during compilation
- Apply explicit pre-pass escaping of `$` followed by digit

---

## Context

Encountered both bugs while restructuring a multi-app monorepo (two Laravel 12 apps in sibling subdirs, each with its own `boost.json`). Bug 1 fired during initial setup of the second app. Bug 2 surfaced when re-running `boost:update` after editing `.ai/guidelines/project.md`, then observing the corrupted regen of `CLAUDE.md`. Reverted both runs and have avoided running `boost:update` since.

Happy to test a candidate fix on our setup.

Thanks for Boost!

impruthvi:
## Summary

- `selectThirdPartyPackages()` and `selectIntegrations()` both called `multiselect()` without checking `isInteractive()`, meaning they could block when `UpdateCommand` calls `InstallCommand` with `--no-interaction` in a terminal environment (where `stream_isatty(STDIN)` is still `true`)
- Added the same `isInteractive()` guard already present in `selectAgents()` to both methods
- Added `tests/Feature/Console/InstallCommandNonInteractiveTest.php` covering the `UpdateCommand → InstallCommand` non-interactive path

Closes #800

## Test plan

- [X] `vendor/bin/pest` — 801 tests pass
- [X] `vendor/bin/phpstan` — no errors
- [X] `UpdateCommand::handle()` with saved config → calls `InstallCommand` with `--no-interaction` → no multiselect prompt blocks

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -220,14 +220,20 @@ class InstallCommand extends Command
             return collect();
         }
 
+        $defaults = collect($this->config->getPackages())
+            ->filter(fn (string $name) => $packages->has($name))
+            ->values();
+
+        if (! $this->input->isInteractive()) {
+            return $defaults;
+        }
+
         return collect(multiselect(
             label: 'Which third-party AI guidelines/skills would you like to install?',
             options: $packages->mapWithKeys(fn (ThirdPartyPackage $pkg, string $name): array => [
                 $name => $pkg->displayLabel(),
             ])->toArray(),
-            default: collect($this->config->getPackages())
-                ->filter(fn (string $name) => $packages->has($name))
-                ->values(),
+            default: $defaults->all(),
             scroll: 10,
             hint: 'You can add or remove them later by running this command again',
         ));
@@ -253,10 +259,18 @@ class InstallCommand extends Command
             ],
         ])->filter(fn (array $integration): bool => $integration['available']);
 
+        $defaults = $integrations->filter(fn (array $integration): bool => $integration['default'])->keys()->all();
+
+        if (! $this->input->isInteractive()) {
+            $this->selectedBoostFeatures->push(...$defaults);
+
+            return;
+        }
+
         $selected = multiselect(
             label: 'Which integrations would you like to configure for Boost?',
             options: $integrations->map(fn (array $integration): string => $integration['label'])->all(),
-            default: $integrations->filter(fn (array $integration): bool => $integration['default'])->keys()->all(),
+            default: $defaults,
             hint: 'Selected integrations will have their MCP servers or skills automatically configured',
         );
 
@@ -300,6 +314,13 @@ class InstallCommand extends Command
                 ->filter(fn (string $name) => $filteredAgents->has($name))
             )
             ->values();
+
+        if (! $this->input->isInteractive()) {
+            return $defaults
+                ->map(fn (string $name) => $filteredAgents->get($name))
+                ->filter()
+                ->values();
+        }
 
         $selected = multiselect(
             label: 'Which AI agents would you like to configure?',
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -484,7 +484,13 @@ class InstallCommand extends Command
 
     protected function isExplicitFlagMode(): bool
     {
-        return (bool) ($this->option('guidelines') || $this->option('skills') || $this->option('mcp'));
+        if ($this->option('guidelines')) {
+            return true;
+        }
+        if ($this->option('skills')) {
+            return true;
+        }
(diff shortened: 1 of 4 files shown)


pushpak1300:
Whats exactly the issue ? How i can reproduce it ?

impruthvi:
The issue only shows up in clean environments with no AI agents installed (like a fresh Docker container). On a dev machine, agents get auto-detected and pre-fill the defaults, so `required: true` never fails, which is why it's tricky to reproduce locally.

Quick repro:
```bash
docker run --rm -it php:8.3-cli bash -c "
  apt-get install -y zip unzip git curl -qq &&
  curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer -q &&
  composer create-project laravel/laravel /app --quiet &&
  cd /app && composer require laravel/boost --quiet &&
  php artisan boost:install --no-interaction
"
```

You'll get:
```
Laravel\Prompts\Exceptions\NonInteractiveValidationException
Required.
```

What's happening: `selectAgents()` calls `multiselect(required: true)`. When running non-interactively, Laravel Prompts calls `default()` instead of rendering the prompt. If no agents are detected (empty defaults) and `required: true` is set, it throws.

The fix adds an early return before the multiselect when `!$this->input->isInteractive()`, reads saved config directly and skips the prompt entirely. After the fix the command completes cleanly.

impruthvi:
After the fix, the command completes cleanly with no exception:

Screenshot 2026-06-04 at 7 39 24 PM

edjeavons:
I had the same issue when updating Boost v2.4.8 to v2.4.9 and confirm this PR resolved the issue for me.

I'm not in a clean environment though. In my case it's because Gemini support has been prematurely dropped by Boost and the new omission isn't handled. Config in case useful:

```json
{
    "agents": [
        "gemini"
    ],
    "cloud": false,
    "guidelines": true,
    "mcp": true,
    "nightwatch": false,
    "sail": false,
    "skills": [
        "laravel-best-practices",
        "livewire-development",
        "pest-testing",
        "tailwindcss-development"
    ]
}
```

PR #842, opened 2026-06-06
Restrict Pint analysis to PHP files only

### LINKED ISSUE #825: Restrict pint to only running on PHP files
### Boost Version

2.4.8

### Laravel Version

13.11.12

### Agent/IDE Name & Version

Gemini, Kiro & PhpStorm, Antigravity

### PHP Version

8.4.21

### System Info

macOS Tahoe with Herd, PhpStorm & Antigravity

### Description

Pint linting is triggered when only updating CSS or JS files, in addition to PHP files. While the current wording says to run pint "if you have modified PHP issues" there needs to be an explicit line added to not run if no php files are modified. 

This version of the prompt has corrected the issue in my use cases:

```
=== pint/core rules ===

# Laravel Pint Code Formatter

- Only run Pint when you have modified PHP files.
- If you have modified any PHP files, run `vendor/bin/pint --dirty --format agent` before finalizing changes to format only the modified files. Do not run Pint if no PHP files were modified.
```

### Steps To Reproduce

Request amends that only involve alteration to non PHP files, and the agent will still trigger Pint after making alterations.

bajelanmehran:
This PR restricts Pint execution to PHP files only.

Changes:
- Filter non-PHP files before passing paths to Pint
- Added tests covering markdown and json files

Closes #825

### DIFF
--- a/.ai/pint/core.blade.php
+++ b/.ai/pint/core.blade.php
@@ -3,6 +3,7 @@
 @endphp
 # Laravel Pint Code Formatter
 
+- Only run Pint when you have modified PHP files. Do not run Pint if no PHP files were modified.
 @if($assist->supportsPintAgentFormatter())
 - If you have modified any PHP files, you must run `{{ $assist->binCommand('pint') }} --dirty --format agent` before finalizing changes to ensure your code matches the project's expected style.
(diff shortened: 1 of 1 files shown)


pushpak1300:
I think this can be added to your pint config manually. https://laravel.com/docs/13.x/pint#excluding-files-or-folders

PR #845, opened 2026-06-08
feat(pint): add --dirty flag for faster formatting

calebdw:
Hello!

This tells the agents to use the `--parallel` option with the `--dirty` option so that it completes much faster.

Thanks!

### DIFF
--- a/.ai/pint/core.blade.php
+++ b/.ai/pint/core.blade.php
@@ -4,9 +4,9 @@
 # Laravel Pint Code Formatter
 
 @if($assist->supportsPintAgentFormatter())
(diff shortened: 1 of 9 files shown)


pushpak1300:
Removed `--parallel` from the guideline. As of Pint v1.29.3, the flag is still marked **Experimental** in the source (`'Runs the linter in parallel (Experimental)'`). In an AI agent context, an experimental flag that silently fails or skips files would cause formatting to appear successful when it isn't — so keeping just `--dirty` is safer until Pint stabilizes the feature.

calebdw:
@nunomaduro, how long will this be in experimental? I've been using it ever since it came out with no issues

PR #849, opened 2026-06-11
Add MCP server icon metadata

sneycampos:
## What changed

Adds icon metadata to the Laravel Boost MCP server so MCP clients can display a branded Boost icon in source/tool surfaces. Currently i'm using and tested in Codex App (OSX)

## Why

Clients that surface MCP server sources can read `serverInfo.icons` from the MCP initialize response. Boost already has branding, but the MCP server currently does not expose an icon, so clients fall back to a generic MCP/source icon.

The SVG is embedded as a data URI because the existing `art/` assets are excluded from package exports.

## Validation

- `vendor/bin/pint src/Mcp/Boost.php tests/Feature/Mcp/BoostServerTest.php --test`
- `vendor/bin/pest tests/Feature/Mcp`

## Example
image


pushpak1300:
Thanks.

PR #856, opened 2026-06-19
fix: make MCP tool timeout configurable

### LINKED ISSUE #817: Tool Timeout is not configurable
### Boost Version

2.4.7

### Laravel Version

13.9

### Agent/IDE Name & Version

Codex

### PHP Version

8.5

### System Info

Ubuntu 22.04

### Description

I have mcp setup in `config.toml` as 
```
[mcp_servers.laravel-boost]
command = "php"
args = ["artisan", "boost:mcp"]
tool_timeout_sec=600

```
Now ideally when an agent calls any of the mcp tools it should run upto 600 seconds.

However, the code at https://github.com/laravel/boost/blob/00ad91567443ddb3d380cc737f625e9264fe0778/src/Mcp/ToolExecutor.php#L70
timeouts any call above 180 seconds . Ideally, we should either be able to set this from boost.php config file or not have it all and let the agent/mcp config decide.  I have some long running jobs for Tinker tool or some long database query tool calls and they timeout. 

### Steps To Reproduce

Ask AI to run any tool like tinker for greater than 3 mins with sleep or something like that and it will timeout at exactly 180seconds.
Change the line in code to higher value than 'tool_timeout_sec' makes it respect 'tool_timeout_sec'

jstar0:
Fixes #817.

## Summary

This makes Boost's internal MCP tool subprocess timeout configurable instead of always using the previous hardcoded 180 second default.

## Changes

- Added a `boost.mcp.tool_timeout` config value backed by `BOOST_MCP_TOOL_TIMEOUT`.
- Kept explicit per-call `timeout` arguments as the highest-precedence override.
- Preserved the existing 1 to 600 second timeout clamp.
- Added a regression test for the configured default timeout path.

## Verification

```bash
composer test
composer test:lint
composer test:types
```

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -32,21 +32,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Boost MCP
-    |--------------------------------------------------------------------------
-    |
-    | This option controls how many seconds Boost will wait for an MCP tool
-    | subprocess to finish before timing out. Individual tool calls may pass a
-    | timeout argument to override this default for that execution.
-    |
-    */
-
-    'mcp' => [
-        'tool_timeout' => env('BOOST_MCP_TOOL_TIMEOUT', 180),
-    ],
-
-    /*
-    |--------------------------------------------------------------------------
     | Boost Executables Paths
     |--------------------------------------------------------------------------
     |
--- a/config/boost.php
+++ b/config/boost.php
@@ -32,6 +32,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Boost MCP
+    |--------------------------------------------------------------------------
+    |
+    | This option controls how many seconds Boost will wait for an MCP tool
+    | subprocess to finish before timing out. Individual tool calls may pass a
+    | timeout argument to override this default for that execution.
+    |
+    */
+
+    'mcp' => [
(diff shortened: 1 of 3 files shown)


pushpak1300:
A couple of follow-up changes on top of the original commit:

**Removed the 600s upper cap.** `getTimeout()` previously did `max(1, min(600, $timeout))`, so even with the config in place anyone setting a value above 600 (e.g. `BOOST_MCP_TOOL_TIMEOUT=900` or `timeout: 900`) was silently clamped back to 600. That's the same surprise the issue was reporting, one level up — the reporter's request was explicitly to *let the agent/mcp config decide*. It's now `max(1, $timeout)`: only a 1s lower bound is enforced, no ceiling.

**Did not publish the `mcp.tool_timeout` key in `config/boost.php`.** The code still reads `config('boost.mcp.tool_timeout')`, so the option works, but it's niche enough that it doesn't need to ship in the default published config and clutter it for everyone. Override paths remain: pass a per-call `timeout` argument, or add the key to your own published config. Default stays 180 when neither is set.

Resolution order is now: per-call `timeout` arg → `config('boost.mcp.tool_timeout')` → `180`.

pushpak1300 (approved):

PR #852, opened 2026-06-12
Add committed project rules with a record-rule MCP tool

pushpak1300:
Currently agents start every session cold. They rediscover the same decisions, hit the same gotchas, and lean on per-vendor "memory" that is personal, machine-local, and never shared with the team or other agents. The knowledge that makes a returning teammate useful lives in people's heads and scrolls out of context.

This adds a small, shared, committed rule store in `.ai/rules/`, grouped by file area, that every agent and teammate can read. One MCP tool records them:

```
record-rule(
  glob: "app/Http/Controllers/**",
  title: "Extend BaseController for tenant scoping",
  note: "New controllers must extend App\Http\Controllers\BaseController."
)
```

That appends to a committed area file with `paths:` frontmatter:

```markdown
---
paths:
  - "app/Http/Controllers/**"
---

# Controllers

## Extend BaseController for tenant scoping
New controllers must extend App\Http\Controllers\BaseController.
```

Reading is just files. The guideline tells agents to check `.ai/rules` (via `@.ai/rules/index.md` and grep) before they plan or change code, and rule files are committed and reviewed in PRs like code. Set `BOOST_RULES_ENABLED=false` to remove the tool entirely.

A few deliberate choices:

- Committed markdown, not behind MCP. The repo is the source of truth, which is what makes rules team-shared, reviewable, and versioned. MCP is just how agents write them.
- No type tag and no search tool. Retrieval is the agent's native grep plus the harness loading area files, so the repository is now just a writer/parser. Keeps the surface small.
- `paths:` frontmatter to line up with the rule formats agents already understand (Claude `paths:`, Cursor and Amp `globs:`) for a future sync step.

PR #857, opened 2026-06-19
Add a built-in Laravel Boost guidance skill

dbpolito:
When an agent is asked to learn something from a session, remember a project preference, or update guidelines, it often edits generated files like `AGENTS.md` or `CLAUDE.md` directly.

In Boost projects, those files are generated outputs. The real source of truth is `.ai/guidelines`, and changes need to be synced with `{{ $assist->artisanCommand('boost:update') }}`. I often end up adding a local skill like this to teach the agent the correct Boost workflow.

This adds a built-in `laravel-boost` skill so agents know how Boost-managed guidance works by default:
- Add custom project guidance to `.ai/guidelines`
- Do not edit generated files like `AGENTS.md` or `CLAUDE.md`
- Run `boost:update` after guideline or skill changes

It also makes core Boost skills discoverable so this guidance is available whenever Boost skills are installed.


pushpak1300:
Hey, I don’t think we need complete skill for this.  A simple instruction in the core file could also work but I won’t add it now because of the new feature we’re adding to Boost which solves this specific problem. https://github.com/laravel/boost/pull/852

PR #859, opened 2026-06-24
Testing with interactive ptys

ohnotnow:
Hi there!

I was trying to work on another PR and noticed that running `pest` with a regular interactive shell would just hang forever.  I _think_ these changes fix the same pattern across a few tests - they do locally for me - but I've not done a full sweep of macos vs Debian vs Rocky vs Fedora vs ...  But on macos and debian running `pest` now passes.


pushpak1300:
thanks.

Issue #851, opened 2026-06-12
boost:install with opencode option creates opencode.json instead of updating existing opencode.jsonc


SantosVilanculos:
## Description

When running `php artisan boost:install` and selecting `opencode` as the AI agent, if an `opencode.jsonc` file already exists in the project, the command creates a new `opencode.json` file instead of updating the existing `opencode.jsonc`.

## Steps to Reproduce

1. Have an existing project with an `opencode.jsonc` file
2. Run `php artisan boost:install`
3. Select `opencode` from the agent options
4. Observe that `opencode.json` is created (instead of updating `.jsonc`)

## Expected Behavior

If `opencode.jsonc` already exists, it should be updated in place. If neither exists, `opencode.json` should be created.

## Actual Behavior

A new `opencode.json` (without the `c` suffix) is created regardless of whether `opencode.jsonc` exists.

## Environment

- Laravel Boost version: latest
- OS: Linux

PR #862, opened 2026-06-27
fix: prefer opencode.jsonc when updating existing configuration

alaminfirdows:
## Description

When running `php artisan boost:install` with the OpenCode agent selected, if an `opencode.jsonc` file already exists in the project, the command now updates it in place instead of creating a new `opencode.json` file.

## Why This Fix

The issue reported in #851 showed that the install command always created `opencode.json` regardless of whether `opencode.jsonc` already existed. This is problematic for users who have already configured OpenCode with a `.jsonc` file (which supports comments).

## Changes Made

1. Modified `OpenCode::projectDetectionConfig()` to detect both `opencode.jsonc` and `opencode.json` files
2. Updated `OpenCode::mcpConfigPath()` to prefer `opencode.jsonc` when it exists, falling back to `opencode.json` as the default
3. Added comprehensive tests to verify the new behavior

## Backwards Compatibility

This change is fully backwards compatible:
- Projects with existing `opencode.json` continue to work
- Projects with `opencode.jsonc` are now handled correctly
- New projects default to creating `opencode.json` as before

### DIFF
--- a/src/Install/Agents/OpenCode.php
+++ b/src/Install/Agents/OpenCode.php
@@ -38,7 +38,7 @@ class OpenCode extends Agent implements SupportsGuidelines, SupportsMcp, Support
     public function projectDetectionConfig(): array
     {
         return [
-            'files' => ['opencode.json'],
+            'files' => ['opencode.json', 'opencode.jsonc'],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


ghost (commented):
I

PR #869, opened 2026-07-03
feat(agents): add MCP configuration support to Antigravity

imKenjo18:
Adds MCP server support to the **Antigravity** agent, allowing the `boost:install` command to automatically configure the Laravel Boost MCP server inside the project's `.agents/mcp_config.json` file.

Also removes the check for `GEMINI.md` from the project detection config, as the default guidelines path for the agent is `AGENTS.md`.

I am testing on Windows, and I've noted that `artisan` often requires absolute paths for it to be found.

> [!NOTE]
> I would appreciate feedback or testing from macOS/Linux users to verify that this works out of the box.


pushpak1300:
Not working in mac.

pushpak1300:
can you verify if it works now ?

imKenjo18:
@pushpak1300

Tested using the default mcp config (no absolute paths), apparently works fine for Antigravity CLI but not in IDE. I don't know for Antigravity 2.0. Again, this is only for Windows. Laravel Boost version is 2.4.13. Should I test with 2.5.0?

pushpak1300:
yes same for me it does not work in the antigravity IDE

imKenjo18:
How about for the CLI? Would it be fine to add it if CLI works but not for IDE or 2.0?

imKenjo18:
Would love this added cuz it's too tiring to look for the mcp config json and copy paste it everytime. Although, this could confuse some users seeing the MCP config but doesn't work.

PR #874, opened 2026-07-08
Support `database-query` tool for MongoDB users.

paulinevos:
The `database-query` tool only supports SQL queries currently. For MongoDB users, the tool attempts to pass SQL (which of course fails), then may try to pass MQL in some cases, before resorting to Tinker.

This has a higher cost, especially in tool calls. This PR introduces support for running read-only MQL commands through the tool. Please see below for a comparison.

> [!NOTE]
> One caveat is that a `database-connections` call will be done if the driver isn't already in context. Assuming most Boost users are not using a `mongodb` driver, we can also opt to only do the `database-connections` call if the first (SQL) attempt fails. This means 2 extra initial tool calls for MongoDB users, but not the one extra initial tool call for most other users. Lmk if you'd like to see that change.

## Before this change:

Screenshot 2026-07-08 at 12 19 58

*Token cost:* ~0.44
 *Tool calls:* 6

## After the

Screenshot 2026-07-08 at 12 23 10
 change:

*Token cost:* ~0.40
 *Boost tool calls:* 2

### DIFF
--- a/.ai/boost/core.blade.php
+++ b/.ai/boost/core.blade.php
@@ -3,7 +3,7 @@
 
 ## Tools
 - Laravel Boost is an MCP server with tools designed specifically for this application. Prefer Boost tools over manual alternatives like shell commands or file reads.
-- Use `database-query` to run read-only queries against the database instead of writing raw SQL in tinker.
+- Use `database-query` to run read-only queries against the database instead of writing raw SQL or MQL in tinker.
 - Use `database-schema` to inspect table structure before writing migrations or models.
 - Use `get-absolute-url` to resolve the correct scheme, domain, and port for project URLs. Always use this before sharing a URL with the user.
 @if (config('boost.browser_logs', false) !== false || config('boost.browser_logs_watcher', true) !== false)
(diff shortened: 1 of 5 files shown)


pushpak1300:
Thanks for your pull request!

paulinevos:
Hi @pushpak1300, that's understandable. And good to know, as we had a few support improvements we wanted to contribute.

For now, I think the best way forward is to create our own version of boost's database tools and suggest them under our package's boost guidelines. Thanks!

pushpak1300:
You can extend all these things as a package or add individual tools in boost itself. This isn’t documented but it’s something you can use to add additional tools and guidelines. just ask agent to build this in your own app.

paulinevos:
~~@pushpak1300 Sorry last question, just to check if I correctly understood this part:~~

~~> add individual tools in boost itself.~~

~~Like, add a guideline file in our third-party package telling Boost to add our own tool during `art boost:install`?~~

Nevermind, found it :) Register the tools in the service provider

PR #875, opened 2026-07-08
feat: add configurable browser log levels

### LINKED ISSUE #289: Add configurability to browser log level
Some projects have JS warnings that are filling their browser logs because they aren't able to prioritize fixing them right now.

They also get sent to Ray and can be pretty noisy showing up on every page load.

How could we add configurability to browser log levels, so people could make it only show errors, say?

JaiveerChavda:
## Summary

This PR adds configuration for Laravel Boost browser log levels, allowing projects to choose which browser console events should be captured by Boost.

By default, the existing behavior remains unchanged. When `boost.browser_log_levels` is configured, Boost only captures and posts the configured browser log types.

Closes #289.

## Problem

Some projects emit noisy browser warnings on every page load. Because Boost captures supported browser console output, those warnings can also be forwarded into downstream debugging tools, making real issues harder to spot.

This PR adds a way to restrict what Boost captures, for example allowing only `console.error()` logs while ignoring `console.warn()`, `console.info()`, and `console.log()`.

## What Changed

* Added a `browser_log_levels` configuration option to the Boost config
* Preserved the existing default browser logging behavior
* Applied filtering in `BrowserLogger` before browser logs are queued and posted
* Normalized warning handling so configured `warning` matches browser `console.warn()`

## Configuration Example

```env
BOOST_BROWSER_LOG_LEVELS=error,warning
```

## Behavior

Developers can control the browser log severity captured by Boost through the `.env` value:

```env
BOOST_BROWSER_LOG_LEVELS=warning
```

| .env value | Captured browser log events |
|--------|--------|
| `error` | error |
| `warning` | warning, error |
| `info` | info, warning, error |
| `debug` | log, info, warning, error, table | 

- `BOOST_BROWSER_LOG_LEVELS=error` captures only browser error-level events
- `BOOST_BROWSER_LOG_LEVELS=warning` captures browser warnings and errors
- `BOOST_BROWSER_LOG_LEVELS=info` captures browser info, warnings, and errors
- `BOOST_BROWSER_LOG_LEVELS=debug` captures all supported browser console events, including console.log(), console.info(), console.warn(), console.error(), and console.table()
- Leaving the value unset preserves the current default behavior

## Notes

* Default behavior is preserved for existing applications
* This change focuses on explicit allowed log-level filtering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,23 +18,6 @@ Laravel Boost accelerates AI-assisted development by providing the essential con
 
 Documentation for Laravel Boost can be found on the [Laravel website](https://laravel.com/docs/boost).
 
-## Browser Log Levels
-
-Boost captures browser console output to provide debugging context. Use `BOOST_BROWSER_LOG_LEVELS` to reduce noisy logs:
-
-```env
-BOOST_BROWSER_LOG_LEVELS=warning
-```
-
-| Value | Captured browser events |
-| --- | --- |
-| `error` | `console.error()` and browser errors |
-| `warning` | warnings and errors |
-| `info` | info, warnings, and errors |
-| `debug` | `console.log()`, `console.debug()`, `console.info()`, `console.warn()`, `console.error()`, and `console.table()` |
-
-The default is `debug`, which preserves Boost's existing behavior by capturing all supported browser events. An empty value also preserves that default. To disable browser logging entirely, set `BOOST_BROWSER_LOGS_WATCHER=false`.
-
 ## Contributing
 
 Thank you for considering contributing to Boost! The contribution guide can be found in the [Laravel documentation](https://laravel.com/docs/contributions).
--- a/README.md
+++ b/README.md
@@ -18,6 +18,23 @@ Laravel Boost accelerates AI-assisted development by providing the essential con
 
 Documentation for Laravel Boost can be found on the [Laravel website](https://laravel.com/docs/boost).
 
+## Browser Log Levels
+
+Boost captures browser console output to provide debugging context. Use `BOOST_BROWSER_LOG_LEVELS` to reduce noisy logs:
+
+```env
+BOOST_BROWSER_LOG_LEVELS=warning
+```
+
+| Value | Captured browser events |
+| --- | --- |
+| `error` | `console.error()` and browser errors |
+| `warning` | warnings and errors |
+| `info` | info, warnings, and errors |
+| `debug` | `console.log()`, `console.debug()`, `console.info()`, `console.warn()`, `console.error()`, and `console.table()` |
+
+The default is `debug`, which preserves Boost's existing behavior by capturing all supported browser events. An empty value also preserves that default. To disable browser logging entirely, set `BOOST_BROWSER_LOGS_WATCHER=false`.
+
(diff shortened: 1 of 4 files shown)


pushpak1300 (commented):
AI Review.

🐞 Bugs (4 distinct, all CONFIRMED)

1. Existing users lose ALL browser logging after upgrade (cached config → null) — CONFIRMED

src/Services/BrowserLogger.php:23, config/boost.php

$levels = json_encode(config('boost.browser_log_levels')) reads with no default. An existing user who published config/boost.php before this feature and ran php artisan config:cache has no browser_log_levels key. Because config is cached, mergeConfigFrom is skipped, so the call returns null → the script emits let levels = null; → shouldCapture's !Array.isArray(levels) guard returns false for every call. All browser logging silently stops on upgrade until they re-publish and re-cache.

Fix: give the read a safe default and, more importantly, make null/empty mean "capture everything" (see design note): config('boost.browser_log_levels') ?? [...defaults].

2. console.table dropped by default — the "unchanged default" claim is false — CONFIRMED

config/boost.php:79, src/Services/BrowserLogger.php:40

Default is error,warning,info,log. severityMap keys are error/warning/info/debug — log isn't a key, so it falls to ?? [level] = ['log']. The default expanded set is {error,warning,info,log} — table is absent. console.table() was captured unconditionally before this PR; now it's silently dropped, and the only value that re-enables it is debug, which the config comment never documents.

Fix: include table in the default set (or make debug the default threshold).

3. Empty BOOST_BROWSER_LOG_LEVELS= disables everything, including uncaught errors — CONFIRMED

config/boost.php:77, src/Services/BrowserLogger.php:82

array_values(array_filter(array_map('trim', explode(',', env(...))))): setting the var to empty (or ,,) makes env() return '' (the default only applies when unset) → []. Then shouldCapture short-circuits false for every type, and all three error paths (window.onerror, error listener, unhandledrejection) bail — so even uncaught errors, previously always captured, stop. A developer trying to "quiet the logs" with an empty value silently blinds Boost entirely.

Fix: treat empty/[] as "capture everything," not "capture nothing."

4. window.onerror guard swallows the site's pre-existing error handler — CONFIRMED

src/Services/BrowserLogger.php:151–155

The new if (!shouldCapture('error')) { return false; } sits at the top of boostErrorHandler, returning before the delegation block at lines 181–183 that chains to a pre-existing window.onerror (e.g. Sentry/Bugsnag). When error capture is off (e.g. BOOST_BROWSER_LOG_LEVELS=log, or the empty/null cases above), the site's own error reporting is permanently swallowed — Boosnrelated tooling.

Fix: gate only the logQueue.pusLet control fall through to theoriginalOnError delegation rega

JaiveerChavda:
Latest two commits address the AI-reported issues above.

With these changes, if BOOST_BROWSER_LOG_LEVELS is not set in .env, or if the package config is still using the default cached value, Boost preserves the previous behavior and captures all browser console events, including console.table()

pushpak1300:
Thanks.

JaiveerChavda:
Learnt so much from you're refactoring, will keep this in mind for next PR :)

PR #876, opened 2026-07-09
Add path-scoped guideline rules

pushpak1300:
Currently every bundled guideline gets inlined into the agent's context in full, on every request, whether or not the agent is touching a file that guideline actually applies to. Model conventions, Inertia conventions, migration conventions, all loaded up front even when you're just editing a controller. That's a lot of dead weight sitting in context on every single request.

This PR lets guideline authors scope a section to the paths it actually applies to:

```blade
# Do Things the Laravel Way

- Use Artisan `make:` commands to create new files.

@scoped(['app/Models/**'])
### Model Creation

- When creating new models, create useful factories and seeders for them too.
@endscoped

@scoped(['app/Http/**', 'routes/**'])
### APIs & Eloquent Resources

- Default to Eloquent API Resources and API versioning.
@endscoped
```

At install time, each `@scoped` block is pulled out of the inline guideline and written to its own file under `.ai/rules/boost/`, tagged with `paths:` frontmatter matching its globs. A generated `.ai/rules/index.md` maps globs to rule files, so agents can check "does anything apply to the file I'm about to touch" instead of carrying every guideline up front. Content outside a `@scoped` block keeps rendering inline like today.

Also moves the Inertia, API resource, and migration guidelines behind scoped blocks as the first real usage.


src/Concerns/BuildsGuidelineAssist.php
```diff
@@ -0,0 +1,21 @@
...
+ * @property Roster $roster
+ * @property GuidelineConfig $config
+ */
+trait BuildsGuidelineAssist
+{
+    protected function buildGuidelineAssist(): GuidelineAssist
+    {
+        return new GuidelineAssist($this->roster, $this->config);
+    }
+}
```
TitasGailius: Do we need a separate trait for this one-line factory?

Both composers still implement `getGuidelineAssist()` just to delegate to it. Would it be simpler to declare `getGuidelineAssist()` as an abstract method on `RendersBladeGuidelines`?

```php
abstract protected function getGuidelineAssist(): GuidelineAssist;
```

src/Install/GuidelineComposer.php
```diff
@@ -244,12 +273,12 @@ protected function getThirdPartyGuidelines(): Collection
         }
 
         return $guidelines->filter(
-            fn (mixed $guideline, string $name): bool => in_array($name, $this->config->aiGuidelines, true),
+            fn (mixed $guideline, string $key): bool => in_array($packageForKey[$key] ?? $key, $this->config->aiGuidelines, true),
```
TitasGailius: Instead of maintaining a separate `$packageForKey` array solely to recover the package name from each key, could we derive it directly?

```php
Str::beforeLast($key, '/')
```

This makes the key format explicit and avoids keeping a separate array.

src/Install/RuleComposer.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    protected function uniqueSlug(array $paths, array $taken): string
+    {
+        $lastSegments = collect($paths)
+            ->map(fn (string $glob): array => Str::of($glob)
+                ->explode('/')
+                ->filter(fn (string $segment): bool => filled($segment) && ! str_contains($segment, '*') && ! str_contains($segment, '.'))
+                ->values()
+                ->all())
+            ->map(fn (array $segments): string => (string) end($segments))
```
TitasGailius: Could we use the collection's `last()` method here instead of converting it to an array and calling `end()`?


```php
    ->map(fn (string $glob): ?string => Str::of($glob)
        ->explode('/')
        ->filter(fn (string $segment): bool => filled($segment) && ! str_contains($segment, '*') && !str_contains($segment, '.'))
        ->last())
```

src/Install/RuleComposer.php
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * @param  array<int, string>  $paths
+     * @param  array<int, string>  $taken
+     */
+    protected function uniqueSlug(array $paths, array $taken): string
+    {
+        $lastSegments = collect($paths)
+            ->map(fn (string $glob): array => Str::of($glob)
+                ->explode('/')
+                ->filter(fn (string $segment): bool => filled($segment) && ! str_contains($segment, '*') && ! str_contains($segment, '.'))
```
TitasGailius: I think we can combine both negative `str_contains` into `! Str::contains($segment, ['*', '.'])`.

src/Console/InstallCommand.php
```diff
@@ -381,6 +389,33 @@ protected function installGuidelines(): void
...
+
+            $composer->withoutRuleExtraction();
+
+            $this->warn('Could not write path-scoped rules to .ai/rules/boost — keeping them inline in the guidelines instead.');
+
+            return;
+        }
+
+        if ($written !== []) {
+            $this->info(sprintf('Extracted %d path-scoped rule file%s to .ai/rules/boost', count($written), count($written) === 1 ? '' : 's'));
```
TitasGailius: Super optional here but we could use `Str::plural('rule file', $count)`.

TitasGailius (approved):
Great job @pushpak1300! This looks great overall. I didn't find any critical issues, just a few small suggestions around readability.

src/Concerns/BuildsGuidelineAssist.php
pushpak1300: yeah make sense. Thanks.

joetannenbaum (approved):

PR #882, opened 2026-07-16
Make third-party guideline discovery the default in boost:update

### LINKED ISSUE #879: `boost:update` ignores new third-party guidelines by default
Currently the only way to get new third-party guidelines is to run `boost:update --discover`, which I think few users will realistically do. 

I would propose `artisan boost:update` to take new third-party guidelines into account by default, and `--no-discover` as an explicit flag. Especially since discovery prompts the user for newly found guidelines.

paulinevos:
Closes #879

New third-party guidelines are only surfaced when running `boost:update --discover`. Few users will realistically use that flag, so in practice new third-party guidelines never get offered on update.

Since discovery already just *prompts* the user before including any newly found guidelines (and is skipped entirely in non-interactive mode), it's safe to run by default.

### What changed

- `boost:update` now runs discovery by default.
- `--no-discover` added as an explicit opt-out.
- `--discover` stays accepted so existing usage keeps working.

### DIFF
--- a/src/Console/UpdateCommand.php
+++ b/src/Console/UpdateCommand.php
@@ -17,7 +17,8 @@ class UpdateCommand extends Command
 {
     /** @var string */
     protected $signature = 'boost:update
-        {--discover : Discover and prompt for newly available guidelines and skills}
+        {--discover : Discover and prompt for newly available guidelines and skills (default)}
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

pushpak1300 (approved):

PR #885, opened 2026-07-17
Add infer-conventions skill to record app conventions

pushpak1300:
Currently a new Boost user starts with an empty `.ai/rules/` and has to hand write the conventions their codebase already follows. Boost can record rules one at a time with the `record-rule` MCP tool, but nothing bootstraps them from how the app is actually written.

This PR ships an `infer-conventions` skill. Instead of a deterministic command, it drives the agent to sweep the app across ~49 Laravel convention dimensions (validation, models, architecture, testing, frontend, database), discover the house patterns a fixed detector can't name (Action classes, query objects, module layout, deliberate non-patterns), and record the real decisions as path-scoped rules via `record-rule`.

Run the skill and it maps your `app/` tree, works each detection, shows the candidates with evidence, and records the ones you approve. `boost:install` now nudges you to run it at the end.

### DIFF
--- a/.ai/boost/skill/infer-conventions/references/checklist.blade.php
+++ b/.ai/boost/skill/infer-conventions/references/checklist.blade.php
@@ -3,7 +3,7 @@
 @endphp
 # Detection Checklist
 
-Every dimension here is a genuine fork: Laravel offers two or more valid approaches, the app's choice changes what the next agent writes, and no tool can pick for you. Left out on purpose: pure formatting (Pint owns it), any form a Rector rule rewrites to one canonical shape (`$casts` to `casts()`, `$fillable` to attributes, pipe-string rules to arrays, named to anonymous migrations, `$signature` to `#[Signature]`), and framework defaults any agent writes unprompted (`ShouldQueue` jobs, relation return types, `HasFactory`).
+Every dimension here is a genuine fork: Laravel offers two or more valid approaches, the app's choice changes what the next agent writes, and no active project tool can pick for you. Left out on purpose: pure formatting (Pint owns it), any form an installed and enabled Rector rule rewrites to one canonical shape (`$casts` to `casts()`, `$fillable` to attributes, pipe-string rules to arrays, named to anonymous migrations, `$signature` to `#[Signature]`), and framework defaults any agent writes unprompted (`ShouldQueue` jobs, relation return types, `HasFactory`).
 
 Each item gives the fork, then a hint (a grep or dir to spot which side the app takes). Hints are only a start. Read the matched files, never record on a raw count. Apply the ground rules to every verdict: a consistent choice that is a default or a tool's target form is not a pattern. Rows tagged (architecture) are the highest-signal, so record presence and deliberate absence.
 
@@ -58,8 +58,8 @@ Each item gives the fork, then a hint (a grep or dir to spot which side the app 
     - Hint: grep `function scope` / `#[Scope]` in models; `ls app/*/Builders`.
 19. Model events: observers (`app/Observers`, `#[ObservedBy]`) vs `booted()` closures vs event classes.
     - Hint: `ls app/Observers`; grep `booted`, `::observe`, `#[ObservedBy]`.
-20. Eager-load posture: per-query `->with()` vs model `$with` vs `preventLazyLoading()`.
-    - Hint: grep `protected $with`, `preventLazyLoading` in `app/`.
+20. Eager-load posture: explicit per-query `->with()` vs model-level `$with` defaults. Treat `preventLazyLoading()` separately as a development guard because it can complement either posture.
+    - Hint: grep `protected $with`, `->with(`, and separately `preventLazyLoading` in `app/`.
 
 ## E. Architecture & organization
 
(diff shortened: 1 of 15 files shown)


src/Console/InstallCommand.php
```diff
@@ -92,6 +93,11 @@ public function handle(): int
         $this->discoverEnvironment();
         $this->collectInstallationPreferences();
         $this->performInstallation();
+
+        if ($this->selectedBoostFeatures->contains('skills')) {
```
WendellAdriel: This fires whenever the `skills` feature is selected, so it still shows when `infer-conventions` was excluded via `boost.skills.exclude` or when no skills-capable agent received the skill, pointing the user at a skill they don't have. Let's gate it on `in_array('infer-conventions', $this->installedSkillNames, true)` instead (it already reflects exclusions and defaults to `[]`), and add an install-command test asserting the note shows when the skill is installed and stays silent when it isn't.

tests/Unit/Install/SkillWriterTest.php
```diff
@@ -36,6 +55,82 @@ function cleanupSkillDirectory(string $path): void
...
+    $relativeTarget = '.boost-test-skills-'.uniqid();
+    $absoluteTarget = base_path($relativeTarget);
+
+    writeInferConventions($relativeTarget);
+
+    $skill = file_get_contents($absoluteTarget.'/infer-conventions/SKILL.md');
+    $checklist = file_get_contents($absoluteTarget.'/infer-conventions/references/checklist.md');
+
+    expect($skill)
+        ->not->toContain('app/Livewire/**')
```
WendellAdriel: `app/Livewire/**` never appears in either template in any render branch, so this assertion can never fail and implies the Livewire-installed render includes that glob when it doesn't. Let's drop it.

tests/Unit/Install/SkillWriterTest.php
```diff
@@ -3,8 +3,27 @@
...
 
 use Laravel\Boost\Contracts\SupportsSkills;
+use Laravel\Boost\Install\GuidelineAssist;
+use Laravel\Boost\Install\GuidelineConfig;
 use Laravel\Boost\Install\Skill;
 use Laravel\Boost\Install\SkillWriter;
+use Laravel\Roster\PackageCollection;
+use Laravel\Roster\ProjectManager;
+
+function writeInferConventions(string $relativeTarget): void
```
WendellAdriel: This helper duplicates the body of the first test below and throws away the `SkillWriter::write()` result, so the two render tests never confirm the write succeeded before reading the output files. A failure would surface as a confusing "failed to open stream" instead of a clear assertion. Let's have it return the `int` result, reuse it in the first test, and assert `SkillWriter::SUCCESS` in both render tests.

.ai/boost/skill/infer-conventions/references/checklist.blade.php
```diff
@@ -0,0 +1,152 @@
...
+   - Hint: typed model params in signatures vs `findOrFail(` in controllers; grep `Route::bind`.
+10. Rate limiting: named `RateLimiter::for()` + `throttle:name` vs inline `throttle:60,1`.
+    - Hint: grep `RateLimiter::for` in providers vs `throttle:` in route files.
+
+## C. Authorization
+
+11. Authorization home: Gates (`Gate::define`) vs Policy classes in `app/Policies`.
+    - Hint: `ls app/Policies`; grep `Gate::define` in `app/Providers`.
+12. Authorization call site: `$this->authorize()` / `Gate::authorize()` vs `$user->can()` vs `can` middleware vs `#[Authorize]` vs `@can` in Blade.
+    - Hint: grep `authorize(`, `->can(`, `middleware('can:`, `#[Authorize(`, `@can(`.
```
WendellAdriel, line 43: The literal `@can(`, `@include`, and `@props` tokens only survive Blade rendering because the backtick placeholder in `RendersBladeGuidelines` glues a word character before the `@`, making Blade's `\B@` directive regex skip them. A future edit that puts one of these tokens after a space inside a code span would get compiled and corrupt the rendered checklist, and no test would notice. Let's add a positive assertion that the rendered checklist still contains these literal tokens, or add them to the placeholder map like `@volt`.

src/Install/SkillComposer.php
```diff
@@ -53,12 +53,26 @@ public function skills(): Collection
         $excluded = config('boost.skills.exclude', []);
 
         return $this->skills = collect()
+            ->merge($this->getCoreSkills())
```
WendellAdriel, line 56: With core skills now shipping to every project unconditionally, `boost.skills.exclude` becomes the only opt-out, and it isn't in the published config, README, or docs. Let's add a `skills.exclude` entry to `config/boost.php` or document it in the README so users can discover it.

tests/Feature/Console/ListSkillCommandTest.php
```diff
@@ -6,6 +6,7 @@
 
 beforeEach(function (): void {
     File::deleteDirectory(base_path('.ai/skills'));
+    config(['boost.skills.exclude' => ['infer-conventions']]);
```
WendellAdriel, line 10: It isn't obvious why every test in this file excludes the new skill. Let's add a one-line comment saying this keeps the tests exercising the empty and user-skill states now that a core skill always ships.

WendellAdriel (changes requested):
The PR body says the nudge happens "at the end of an interactive install", but it also shows on non-interactive runs since the check never looks at interactivity. Let's either update the description or add the interactivity check if interactive-only was the intent.

.ai/boost/skill/infer-conventions/references/checklist.blade.php
pushpak1300: make sense.

src/Install/SkillComposer.php
pushpak1300: i think this is intentional. will plan adding it seprately.

WendellAdriel (approved):
LGTM :muscle:

PR #886, opened 2026-07-20
[2.x] Update .gitattributes export ignore entries

jackbayliss:
This ensures the `workbench` folder, `pint.json` and `rector.php` aren't exported, so they dont end up in the end users vendor. I was going through our vendor installs and noticed this. 

image

There could be more to ignore like resources, but did these for now.


pushpak1300:
HYB

PR #889, opened 2026-07-22
Support `extra.laravel-boost` package discovery in `composer.json`

rojtjo:
# Support `extra.laravel-boost` package discovery in composer.json

Boost discovers guidelines by reading `require` and `require-dev` from the application's `composer.json`. Applications that pull packages in through an internal meta-package get nothing, because those packages never appear as a direct requirement.

This PR adds two opt-in keys under `extra.laravel-boost` so an application can tell Boost about packages it can't see.

## Why

We currently maintain multiple applications that share a common package containing about 90% of their functionality. As a result, each application typically requires only templating work and the implementation of a small number of custom features.

This shared package depends on several other packages, many of which provide their own guidelines and/or skills. At present, we must either publish these skills manually or duplicate the relevant requirements in each application’s composer.json file.

These changes allow us to just add the shared package to the `include-packages-from` list to have all related guidelines and/or skills discovered. It also supports adding additional packages explicitly without having to keep the version in sync.

## `packages` — list packages explicitly

```json
{
    "extra": {
        "laravel-boost": {
            "packages": [
                "spatie/laravel-permission",
                "livewire/livewire"
            ]
        }
    }
}
```

Both packages are now treated as if they were listed in `require`, so their guidelines and skills are discovered.

## `include-packages-from` — inherit a package's requirements

Given an internal meta-package that every project in the org requires:

```json
{
    "require": {
        "acme/bundle": "^1.0"
    },
    "extra": {
        "laravel-boost": {
            "include-packages-from": [
                "acme/bundle"
            ]
        }
    }
}
```

And `vendor/acme/bundle/composer.json`:

```json
{
    "require": {
        "php": "^8.4",
        "spatie/laravel-permission": "^6.0",
        "livewire/livewire": "^3.0"
    }
}
```

Boost now sees `spatie/laravel-permission` and `livewire/livewire`. Platform requirements (`php`, `ext-*`, `lib-*`, `composer-*`) are skipped.

Both keys can be used together, and `include-packages-from` accepts multiple sources.

## Indirect first-party packages

`livewire/livewire` and `laravel/mcp` are normally only given guidelines when required directly — this stops every Boost user picking up MCP guidelines because something else pulled it in. Listing one of them under `extra.laravel-boost` is an explicit opt-in, so it now overrides that rule and the guidelines are included even though the package is indirect.

## Error handling

Sources that don't resolve are ignored rather than fatal:

- `include-packages-from` entry not installed in `vendor/`
- source `composer.json` that is missing, malformed, or doesn't decode to an object
- non-array `extra.laravel-boost`, `packages`, `include-packages-from`, or `require`

### DIFF
--- a/src/Install/Concerns/DiscoverPackagePaths.php
+++ b/src/Install/Concerns/DiscoverPackagePaths.php
@@ -62,7 +62,9 @@ trait DiscoverPackagePaths
             }
         }
 
-        return ! $package->isDirect() && in_array($package->name(), $this->mustBeDirect, true);
+        return ! $package->isDirect()
+            && in_array($package->name(), $this->mustBeDirect, true)
+            && ! in_array($package->name(), Composer::extraPackageNames(), true);
     }
 
     /**
--- a/src/Support/Composer.php
+++ b/src/Support/Composer.php
@@ -34,32 +34,122 @@ class Composer
     public static function packagesDirectories(): array
     {
         return collect(static::packages())
-            ->mapWithKeys(fn (string $key, string $package): array => [$package => implode(DIRECTORY_SEPARATOR, [
-                base_path('vendor'),
-                str_replace('/', DIRECTORY_SEPARATOR, $package),
-            ])])
+            ->mapWithKeys(fn (string $key, string $package): array => [$package => self::vendorPath($package)])
             ->filter(fn (string $path): bool => is_dir($path))
             ->toArray();
     }
 
     public static function packages(): array
     {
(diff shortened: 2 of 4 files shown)


pushpak1300:
Thanks for your pull request!

rojtjo:
Let me know if you every reconsider, happy to send in a new PR at that point!

In case anyone else needs this functionality, see [feature-ninja/boost-extra-packages](https://github.com/feature-ninja/boost-extra-packages)

PR #901, opened 2026-07-29
Handle signaled process when detecting test enforcement

### LINKED ISSUE #900: boost:update crashes when internal test listing process is signaled (signal 5)
### Laravel Boost version
v2.4.13

### Laravel version
v13.23.0

### PHP version
8.3.6

### Related packages
- laravel/prompts: v0.3.21
- symfony/process: v7.4.13
- pestphp/pest: v4.7.5

### Operating system
Linux 7.0.0-28-generic (Ubuntu 24.04, x86_64)

### What happened?
Running `php artisan boost:update -n` crashes with:

`Symfony\Component\Process\Exception\ProcessSignaledException`
`The process has been signaled with signal "5".`

Stack trace points to:
- `vendor/laravel/boost/src/Console/InstallCommand.php:169`
- `determineTestEnforcement()`

### Steps to reproduce
1. Install Boost and initialize config:
   - `composer require laravel/boost --dev`
   - `php artisan boost:install`
2. Run:
   - `php artisan boost:update -n -vvv`

### Actual result
`boost:update` fails with `ProcessSignaledException` (signal 5).

### Expected result
`boost:update` should complete, or at least gracefully fallback if internal test-discovery subprocess fails.

### Additional context
`InstallCommand::determineTestEnforcement()` runs `php artisan test --list-tests` via Symfony Process.

In this environment, `php artisan test --list-tests` itself crashes (sandbox / namespace failure) and then returns signal 5:

```text
The setuid sandbox is not running as root.
Failed to move to new namespace: ... Operation not permitted
[FATAL:zygote_host_impl_linux.cc:207] ...
Symfony\Component\Process\Exception\ProcessSignaledException
The process has been signaled with signal "5".
```

Because this subprocess exception is not handled in `determineTestEnforcement()`, `boost:update` aborts.

### Workaround that works
Setting `enforce_tests` to `false` in `config/boost.php` avoids the crash path and allows:

`php artisan boost:update -n`

to complete successfully.

### Suggestion
Handle `ProcessSignaledException` (or non-zero run) in `determineTestEnforcement()` and fallback to `false`, so update can proceed even when `artisan test --list-tests` cannot run in constrained environments.

lazerg:
`determineTestEnforcement()` runs `php artisan test --list-tests` through Symfony Process to decide whether to enable the test enforcement guideline. In sandboxed or otherwise constrained environments that subprocess can be killed by a signal, and the `ProcessSignaledException` it throws is never caught, so `boost:install` and `boost:update` abort. This catches it and falls back to no enforcement, the same safe default already used when no test runner is present.

Fixes #900

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -33,6 +33,7 @@ use Laravel\Boost\Skills\Remote\RemoteSkill;
 use Laravel\Boost\Support\Config;
 use Laravel\Prompts\Terminal;
 use RuntimeException;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
 use Symfony\Component\Process\Process;
 use Throwable;
 
@@ -179,7 +180,12 @@ class InstallCommand extends Command
         }
 
         $process = new Process([PHP_BINARY, 'artisan', 'test', '--list-tests'], base_path());
-        $process->run();
+
+        try {
+            $process->run();
+        } catch (ProcessSignaledException) {
+            return false;
+        }
 
         return Str::of($process->getOutput())
             ->trim()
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks.

PR #902, opened 2026-07-29
Refactor scope methods to use #[Scope] attribute

mathcrln:
A small change to ensure boost recommends using the latest and currently recommended syntax to define [local scopes](https://laravel.com/docs/13.x/eloquent#local-scopes).

### DIFF
--- a/.ai/laravel/skill/laravel-best-practices/rules/eloquent.md
+++ b/.ai/laravel/skill/laravel-best-practices/rules/eloquent.md
@@ -30,7 +30,8 @@ $articles = Article::whereHas('user', function ($q) {
 
 Correct:
 ```php
-public function scopeActive(Builder $query): Builder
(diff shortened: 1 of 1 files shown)


pushpak1300:
thanks

.ai/laravel/skill/laravel-best-practices/rules/eloquent.md
```diff
@@ -30,7 +30,8 @@ $articles = Article::whereHas('user', function ($q) {
 
 Correct:
 ```php
-public function scopeActive(Builder $query): Builder
+#[Scope]
+protected function active(Builder $query): Builder
 {
     return $query->where('verified', true)->whereNotNull('activated_at');
```
siarheipashkevich, line 36: return - is redundant here

PR #905, opened 2026-08-02
Add Laravel LSP registration to boost:install

mischasigtermans:
Follow-up to laravel/agent-skills#26, which was closed as 'this needs to be part of boost', which is _absolutely right_.

This adds Laravel LSP registration to `boost:install` as a fourth feature next to guidelines, skills, and MCP. For Claude Code, the installer writes a project plugin to `<skills path>/laravel-lsp` that registers [laravel/lsp](https://github.com/laravel/lsp) with Claude Code's LSP tool, giving agents go-to-definition and hover on routes, views, config keys, translations, and middleware aliases instead of grepping for them. Verified end-to-end on a fresh application: `php artisan boost:install --lsp`, open Claude Code, and `view('welcome')` resolves to `resources/views/welcome.blade.php`.

If `laravel-lsp` is missing, the installer offers to run `composer global require laravel/lsp` behind an interactive confirm — the only place Boost would touch global state, and it never happens non-interactively. Detection also checks Composer's global bin directories and warns when the binary is not on `PATH`, since agents could not launch it.

Existing installs opt in with `php artisan boost:install --lsp`, which preserves stored configuration. Scoped to Claude Code (`ClaudeCodeLspWriter`); Zed, VS Code, Cursor, and Sublime already bundle the language server through their official extensions, and OpenCode can follow once it can be verified end-to-end. Tests cover the writer, path derivation from a custom skills path, idempotency, the unwritable-directory failure path, and a `--lsp` flag-mode regression. Pint and PHPStan pass.


DarkGhostHunter:
What if I want to use [Laravel CE LSP](https://github.com/mike-bronner/zed-laravel) (which is faster)?

pushpak1300:
> What if I want to use [Laravel CE LSP](https://github.com/mike-bronner/zed-laravel) (which is faster)?

you can override manually.

danielhe4rt:
@pushpak1300 any idea on when it will get shipped?

pushpak1300:
Still need to test it correctly with supported agents.

PR #907, opened 2026-08-02
Prevent `boost:update` from failing when there is nothing to update

ohnotnow:
Currently `boost:update` needs configured agents before it checks whether there is anything to update, so an MCP-only setup (`php artisan boost:install --mcp`) gets `Please set up Boost with [php artisan boost:install] first.` and a non-zero exit code.

This PR reorders the checks in `handle()` so that when guidelines or skills aren't configured the command exits successfully before requiring the agents part.

Example run that shows the current behaviour in a new project that this PR (seems to!) fix:

```sh
$ composer require --dev laravel/boost
... <all fine>

$ php artisan boost:install --mcp
... <all looks ok>

$ php artisan boost:update
Please set up Boost with [php artisan boost:install] first.
```

### DIFF
--- a/src/Console/UpdateCommand.php
+++ b/src/Console/UpdateCommand.php
@@ -23,14 +23,10 @@ class UpdateCommand extends Command
 
     public function handle(Config $config): int
     {
-        if (! $config->isValid() || empty($config->getAgents())) {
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks

PR #908, opened 2026-08-02
Store selected agents when boost:install and feature flags

ohnotnow:
Currently agents you select during `boost:install` are only saved to `boost.json` if the command is run with no feature flags. Run `php artisan boost:install --mcp` and the MCP config is written for each agent, but `boost.json` ends up without an `agents` entry - so the next `boost:update` gives `Please set up Boost with [php artisan boost:install] first.`

This PR also saves the selected agents when feature flags are used.

Example boost.json after running the current `php artisan boost:install --mcp` (no agents key):

```json
{
    "cloud": true,
    "mcp": true,
    "nightwatch": false,
    "sail": false
}
```

The same run with this PR merged:

```json
{
    "agents": [
        "claude_code"
    ],
    "cloud": true,
    "mcp": true,
    "nightwatch": false,
    "sail": false
}
```

(Side note - I think if this gets merged then my previous #907 is fixed as a side effect)

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -504,6 +504,10 @@ class InstallCommand extends Command
             $this->config->setPackages($this->selectedThirdPartyPackages->values()->toArray());
         }
 
+        if ($explicitMode && $this->selectedAgents->isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


pushpak1300:
i think #907 is correct fix.

Closing this in favour of #907

Issue #895, opened 2026-07-26
10 factual bugs in bundled Boost skill docs (laravel-best-practices + mcp/cashier skill content)


jaballer:
## Summary

While reviewing PR [jaballer/kratecms#1181](https://github.com/jaballer/kratecms/pull/1181) (commit `196eb46`), a code review pass (Codex) flagged a batch of factual errors in the Boost-generated skill content that ships with this repo's dependencies. I independently re-verified each item against the actual vendored files and Laravel's official docs before filing. 6 are in `laravel/boost`'s own bundled `laravel-best-practices` skill; 4 are in *other* packages' self-shipped Boost skill content (`resources/boost/skills/`, which — per Boost's own resolution order — takes precedence over `laravel/boost`'s fallback copy). I've kept all 10 in one issue since they were discovered together, with the non-`laravel/boost` ones clearly called out so they can be redirected/transferred as you see fit.

All file paths below are relative to `vendor/laravel/boost/.ai/...` (or the noted package) as installed via Composer.

---

### 1. (P2) `laravel/skill/laravel-best-practices/rules/security.md:124` — misleading framing of `mimes` vs `extensions`

Current text:

> "Validate extension, MIME type, and size. The `mimes` rule checks extensions; use `mimetypes` for actual MIME type validation."

This implies `mimes` is the "just checks the extension" (weak) rule and `mimetypes` is needed for "actual" content-based validation. That's backwards. Per Laravel's own docs, `mimes` is listed as **"MIME Type By File Extension"** — it reads the file's actual contents and guesses its MIME type, then checks that against the extension list. `mimetypes` does the same content-sniffing, just expressed as raw MIME type strings instead of extensions. The rule that actually trusts the **client-supplied** filename extension is `extensions`, which isn't mentioned at all here, and Laravel's docs explicitly warn never to rely on it alone.

A reader following this doc's mental model could come away thinking `mimes` is naive/extension-trusting when it isn't, or reach for `extensions` alone thinking it's equivalent — the opposite of what this "Validate File Uploads" section is trying to teach.

### 2. (P1) `laravel/skill/laravel-best-practices/rules/architecture.md:107` — `lockForUpdate()` example has no surrounding transaction

```php
// Or at query level
$product = Product::where('id', $id)->lockForUpdate()->first();
```

[...]

PR #918, opened 2026-08-06
Wrap lockForUpdate example in a transaction

ulofiai:
## What changed

- wrap the `lockForUpdate()` example in `DB::transaction()`
- keep the protected read and subsequent update work inside the transaction callback

## Why

A pessimistic row lock only protects the full operation when it remains active through the related reads and writes. The previous example ran the locking query by itself, so the lock could be released before the application performed the protected work.

## Validation

- `git diff --check`
- `php -l` on the updated PHP example

Partially addresses #895 (item 2).


pushpak1300:
Thanks.

PR #922, opened 2026-08-07
Ignore the connection table prefix when reading table details

### LINKED ISSUE #920: DatabaseSchema: prefixed connections silently return empty columns, indexes and foreign keys
### Boost Version

v2.5.2

### Laravel Version

v12.65.0

### PHP Version

8.4.3

### Database Driver & Version

MariaDB 12.1.2 (via the `mysql` driver) — but the defect is driver-independent, see below.

### Description

On any connection configured with a non-empty table prefix, the `database-schema` MCP tool returns **every table with empty `columns`, `indexes` and `foreign_keys`**.

It does not raise an error. It reports success with an empty structure, so an agent reading the output concludes the tables genuinely have no columns. That silence is what makes this worth fixing — the tool looks like it worked.

```jsonc
// database-schema  { "filter": "order_info" }   with 'prefix' => 'grs_'
{
  "engine": "mysql",
  "tables": {
    "grs_order_info": { "columns": [], "indexes": [], "foreign_keys": [], ... }
    //                              ^^ actually has 79 columns and 58 indexes
  }
}
```

### Root cause — the prefix is applied twice

`DatabaseSchema::getAllTables()` ([line 176](https://github.com/laravel/boost/blob/main/src/Mcp/Tools/DatabaseSchema.php)) delegates to `SchemaDriverFactory::make($connection)->getTables()`. Every driver queries the catalog directly, so the names come back **fully prefixed** (`grs_order_info`):

- `MySQLSchemaDriver::getTables()` → `information_schema.TABLES`
- `PostgreSQLSchemaDriver::getTables()` → `pg_tables`
- `SQLiteSchemaDriver::getTables()` → `sqlite_master`

Those raw names are then passed straight into the schema builder at four call sites in `DatabaseSchema.php`:

| Line | Call |
|------|------|
| 163 | `Schema::connection($connection)->getColumns($tableName)` |
| 217 | `$schema->getColumns($tableName)` |
| 248 | `Schema::connection($connection)->getIndexes($tableName)` |
| 266 | `Schema::connection($connection)->getForeignKeys($tableName)` |

`Illuminate\Database\Schema\Builder` **prepends the connection prefix itself**, so the effective lookup is `grs_grs_order_info` — a table that does not exist.

`getColumns()` / `getIndexes()` return `[]` for a missing table, and `getTableForeignKeys()` wraps the call in `catch (Exception) { return []; }`. Hence the silent empty result rather than a failure.

### Steps To Reproduce

1. Set a table prefix on the default connection in `config/database.php`:
   ```php
   'mysql' => [ /* ... */ 'prefix' => 'grs_' ],
   ```
2. Have at least one real table, e.g. `grs_order_info`.
3. Call the `database-schema` tool (either `summary` or full mode).
4. Every table is listed with `"columns": []`.

Confirmed via Laravel's own API:

```php
Schema::getColumns('order_info');      // 79  <- correct, builder adds the prefix
Schema::getColumns('grs_order_info');  // 0   <- what Boost passes in
Schema::getTables()[$n]['name'];       // 'grs_order_info'
```

### Suggested fix

Strip the connection prefix before handing a name back to the schema builder. A local patch that resolves it:

```php
/**
 * Strip the connection's table prefix from a raw table name.
 *
 * getAllTables() reads the catalog directly, so it yields fully prefixed
 * names. Schema::getColumns()/getIndexes()/getForeignKeys() then prepend the
 * prefix again, producing a double-prefixed lookup that silently returns
 * empty results on any connection configured with a table prefix.
 */
protected function withoutTablePrefix(?string $connection, string $tableName): string
{
    $prefix = DB::connection($connection)->getTablePrefix();

    if ($prefix !== '' && str_starts_with($tableName, $prefix)) {
        return substr($tableName, strlen($prefix));
    }

    return $tableName;
}
```

…applied at the four call sites above, e.g. `->getColumns($this->withoutTablePrefix($connection, $tableName))`.

Worth noting: the **driver-level** calls in `getTableStructure()` — `$driver->getTriggers($tableName)` and `$driver->getCheckConstraints($tableName)` — query `information_schema` directly and therefore correctly expect the *prefixed* name. They should be left as-is; only the four `Schema::` builder calls need unprefixing.

After the patch, the same request returns `columns=79, indexes=58` for `grs_order_info`, matching `Schema::getColumnListing('order_info')` exactly. Verified in both `summary` and full (`include_column_details`) modes.

### Related

#867 — same prefix-handling area, but a different tool (`database-query`) and a different defect.

lazerg:
On a connection with a table prefix, the `database-schema` tool returned every table with empty `columns`, `indexes` and `foreign_keys`, without any error.

The table names come from the schema drivers, which read the catalog directly and so return them exactly as stored. `Schema::getColumns()`, `getIndexes()` and `getForeignKeys()` then prepend the connection prefix, so the lookup is for a table that does not exist and the result is empty. This runs those calls inside `Connection::withoutTablePrefix()`, the same way `db:table` handles catalog names. The driver-level calls for triggers and check constraints are untouched, since they query the catalog and already expect the stored name.

Fixes #920

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema.php
+++ b/src/Mcp/Tools/DatabaseSchema.php
@@ -152,18 +152,21 @@ class DatabaseSchema extends Tool
     protected function getAllTableColumnTypes(?string $connection, string $filter = ''): array
     {
         $tables = [];
+        $allTables = $this->getAllTables($connection);
 
-        foreach ($this->getAllTables($connection) as $table) {
-            $tableName = is_object($table) ? $table->name : ($table['name'] ?? '');
+        DB::connection($connection)->withoutTablePrefix(function () use ($connection, $allTables, $filter, &$tables): void {
+            foreach ($allTables as $table) {
+                $tableName = is_object($table) ? $table->name : ($table['name'] ?? '');
 
-            if ($filter !== '' && ! str_contains(strtolower($tableName), strtolower($filter))) {
-                continue;
+                if ($filter !== '' && ! str_contains(strtolower($tableName), strtolower($filter))) {
+                    continue;
+                }
+
+                $tables[$tableName] = collect(Schema::connection($connection)->getColumns($tableName))
+                    ->pluck('type', 'name')
+                    ->all();
             }
-
-            $tables[$tableName] = collect(Schema::connection($connection)->getColumns($tableName))
-                ->pluck('type', 'name')
-                ->all();
-        }
+        });
 
         return $tables;
     }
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

PR #925, opened 2026-08-07
feat: Add GitHub-backed guidelines

### LINKED ISSUE #924: [Feature request] Support installing AI guidelines from GitHub repositories
We have 7 Laravel applications at work. We maintain our shared AI skills in a private GitHub repository and install them in each application using `boost:add-skill`. That workflow has worked very well for us.

Recently, we have needed the same workflow for guidelines. These are standing architectural and development conventions that should be loaded upfront across all seven applications, rather than on-demand instructions represented as skills.

Boost currently lets us place these files directly in each application's `.ai/guidelines` directory or distribute them through an installed Composer or npm package. For our use case, that would mean turning Markdown guidance into a private Composer package and either maintaining a private package registry or configuring the repository as a VCS dependency in every application.

That package would contain no PHP library code and would exist only to transport guideline files. We could also copy the files manually or use Git submodules, but then we would be maintaining a custom installation workflow for guidelines while Boost already provides that workflow for skills.

Would you be open to adding a command such as:

```shell
php artisan boost:add-guideline <owner/repository>
```

Ideally, it would follow the same user-facing model as `boost:add-skill`: fetch guidelines from a GitHub repository, allow the user to select which ones to install, and copy them into `.ai/guidelines` for Boost's existing composition process.

If repository paths are preserved, this would also allow teams to share overrides for built-in Boost guidelines. The installed files would behave exactly as if they had been added locally to `.ai/guidelines`; this would not require a new override mechanism.

I am not proposing that Boost fetch remote instructions whenever it runs. The command could explicitly download the files so they remain local, committed, and reviewable, just like remotely installed skills.

I also noticed #760, which proposes tracking the source repository and path of GitHub-installed skills. If that approach is merged, GitHub-installed guidelines would follow the same metadata model. This would preserve their origin and leave room for future update support across both skills and guidelines.

I consider automatic updates or change detection a follow-up concern rather than a requirement for this proposal. The immediate need is simply the ability to install guidelines from GitHub using the same kind of workflow already available for skills.

clvsh:
This PR addresses #924. It allows teams to share guidelines through GitHub by extending the existing GitHub-backed skill flow to guidelines. It provides a lighter alternative to transport-only Composer packages and makes private guidelines easier to distribute through private GitHub repositories without maintaining a private Composer registry.

The shared GitHub fetching logic now lives in a concern, while `GitHubRepository` has moved to `Support` since it is used by both skills and guidelines. Downloaded Markdown guidelines preserve nested paths and work with the existing guideline override system.

the command will only scan the `guidelines/` directory in the repo. I find this to make more sense.

Closes #924


pushpak1300:
Thanks for your pull request!

PR #928, opened 2026-08-09
Correct the mimes and extensions guidance for file upload validation

sulimanbenhalim:
`laravel-best-practices` has the file upload rule backwards:

> The `mimes` rule checks extensions; use `mimetypes` for actual MIME type validation.

per the [docs](https://laravel.com/docs/12.x/validation#rule-mimes), `mimes` "actually validates the MIME type of the file by reading the file's contents and guessing its MIME type". `mimetypes` does the same, just spelled as MIME strings instead of extensions. the rule that trusts the client is `extensions`, which isn't mentioned at all: "You should never rely on validating a file by its user-assigned extension alone."

covers item 1 of #895. code sample was already fine, left it.


pushpak1300:
Thanks.

PR #933, opened 2026-08-11
Ignore commented-out entries when checking for existing MCP servers

### LINKED ISSUE #932: `boost:install --mcp` reports success without installing when the server key appears in a comment
if `laravel-boost` appears in a comment inside `mcpServers`, install does nothing and still reports ✓.

`.mcp.json`:

```json
{
    "mcpServers": {
        // "laravel-boost": { "command": "php", "args": ["artisan","boost:mcp"] }
    }
}
```

`php artisan boost:install --mcp` prints `Claude Code.. ✓` and exits 0. file comes back byte identical, nothing registered.

cause: `FileWriter::serverExistsInContent()` (`src/Install/Mcp/FileWriter.php:176`) regexes the raw slice between the `mcpServers` braces, comments included. so `filterExistingServers()` comes back empty and `injectIntoExistingConfigKey()` hits the `return true` at `:126`.

what does and doesn't trigger it, inside `mcpServers`:

| comment | installed |
|---|---|
| `// "laravel-boost": {...}` | no |
| `// laravel-boost: disabled` | no |
| `// re-enable laravel-boost later` | yes |

json5 path only. plain json goes through `json_decode` and is fine, and codex/toml is fine too since that check anchors `^\[`. it isn't specific to boost's own key either, I got the same result driving `FileWriter` directly with a `nightwatch` key.

commenting a server out and later running install to bring it back seems normal enough, that's where it bit me.

if anyone picks this up, one thing to watch: stripping `//` before the check would also eat `//` inside string values like `"url": "https://x"`, which flips it into a duplicate insert instead. needs to be string aware.

happy to PR it, just wasn't sure which way you'd want the check done.

lazerg:
`boost:install --mcp` reports success but writes nothing when the server key only shows up in a commented-out line inside `mcpServers`. That happens when you disable a server and later run install to bring it back.

`serverExistsInContent()` matches against the raw slice between the `mcpServers` braces, so `// "laravel-boost": {...}` counts as an existing entry and the server is filtered out before injection. The check now runs on the content with comments removed, and quoted strings are left alone so a `//` inside a value like `"url": "https://example.com"` still reads as real config rather than a comment.

Fixes #932

### DIFF
--- a/src/Install/Mcp/FileWriter.php
+++ b/src/Install/Mcp/FileWriter.php
@@ -175,10 +175,26 @@ class FileWriter
 
     protected function serverExistsInContent(string $content, string $serverKey): bool
     {
+        $content = $this->maskUnquotedComments($content);
+
         $quotedPattern = '/["\']'.preg_quote($serverKey, '/').'["\']\\s*:/';
         $unquotedPattern = '/(?<=^|\\s|,|{)'.preg_quote($serverKey, '/').'\\s*:/m';
 
         return preg_match($quotedPattern, $content) || preg_match($unquotedPattern, $content);
+    }
+
+    protected function maskUnquotedComments(string $content): string
+    {
+        // Match quoted strings (keep) or line and block comments (blank out, keeping length and line breaks)
+        $pattern = '/"(?:\\\\.|[^"\\\\])*"|\'(?:\\\\.|[^\'\\\\])*\'|\\/\\/[^\\n]*|\\/\\*[\\s\\S]*?\\*\\//';
+
+        return preg_replace_callback(
+            $pattern,
+            fn (array $matches): string => Str::startsWith($matches[0], ['//', '/*'])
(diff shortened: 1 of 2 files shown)


sulimanbenhalim:
reporter here. pulled the branch and ran it end to end against a real app: original case is fixed both ways (commented entry next to another server, and as the only content), and a real entry still doesn't duplicate.

also checked the things I was worried about when I filed it, all fine: block comments, an apostrophe inside a comment, braces inside a comment, and `//` inside a string value with comments present.

masking comments to spaces rather than stripping them is a nice call, keeps the offsets valid so `findMatchingClosingBrace` gets fixed as well, which I hadn't considered.

fwiw I dropped the mask call from `serverExistsInContent` locally and 2 of the 3 new tests fail, so they're guarding properly.

lazerg:
Thanks for the thorough check, this confirms the fix holds up.

PR #934, opened 2026-08-13
Include `PATCH` in the CSRF rule

MartinCamen:
Require @csrf in all POST/PUT/PATCH/DELETE Blade forms. PATCH forms use method spoofing but still require CSRF protection.

> Anytime you define a "POST", "PUT", "PATCH", or "DELETE" HTML form in your application, you should include a hidden CSRF _token field in the form so that the CSRF protection middleware can validate the request.

https://laravel.com/docs/13.x/csrf#preventing-csrf-requests


pushpak1300:
Thanks.

PR #935, opened 2026-08-13
feat: support third-party NPM package guidelines and skills

calebdw:
## Problem

Third-party **Composer** packages can already ship Boost guidelines and skills (via `vendor/{pkg}/resources/boost/guidelines/` and `resources/boost/skills/`), but third-party **NPM** packages cannot. The discovery methods (`Npm::packagesDirectoriesWithBoostGuidelines()` / `Npm::packagesDirectoriesWithBoostSkills()`) exist but are never called in the third-party resolution paths — only first-party NPM packages (`@laravel/*`, `@inertiajs/*`) are discovered, through the separate Roster-based code path.

## Solution

Wire NPM third-party discovery into the three places that currently only use `Composer::`:

- **`GuidelineComposer::getThirdPartyGuidelines()`** — merge `Npm::packagesDirectoriesWithBoostGuidelines()` (rejecting first-party NPM packages) alongside the existing Composer discovery.
- **`SkillComposer::getThirdPartySkills()`** — merge `Npm::packagesDirectoriesWithBoostSkills()` into the collection before `flatMap`.
- **`ThirdPartyPackage::discover()`** — merge both Composer and NPM directories, rejecting first-party packages from both systems.

## How it works

Any NPM package can now ship guidelines and skills at:

```
node_modules/{package}/resources/boost/guidelines/
node_modules/{package}/resources/boost/skills/
```

These are discovered automatically, the same way `vendor/{package}/resources/boost/` already works for Composer packages.


valorin:
I find this a bit concerning - and the referenced Composer skill adoption feature too. Correct me if I'm missing something here, but wouldn't this allow a malicious package to inject dodgy instructions into Boost, potentially hijacking the user's agent?

It makes sense for Boost to install Laravel's curated skills, but automatically importing skills from 3rd party packages without any approval feels like a recipe for disaster.

calebdw:
You have to purposefully choose to install vendor skills. And supporting 3rd party npm package are no different than 3rd party composer packages

valorin:
Ah, so it prompts to enable each skill? My mistake! Ignore me. 🫣

calebdw:
When you run `boost:install` you choose which ones you want and trust.  From there `boost:update` will update it automatically if the package updates

valorin:
> From there `boost:update` will update it automatically if the package updates

Been thinking more about this, and automatic updates have the vector I was worried about. If someone can hijack a package with boost guidelines/skills, they can push malicious prompts out via the automatic update to any projects that have already approved then. 

It doesn't like the immediate RCE effect like a post-install/update script though, so it's similar to malicious code inside the package. However, malicious code is probably easier to flag and catch than malicious prompts - especially since the big players probably aren't looking for it like they are malicious code.

calebdw:
Thanks!

PR #938, opened 2026-08-16
Send an idempotency key in the retried charge example

sulimanbenhalim:
The retry rule's "Correct" example retries a POST to Stripe's charge endpoint with no idempotency key.

`Http::retry()` resends the request as-is, so if the first attempt lands but the response gets lost, the retry makes a second charge. Stripe's docs are pretty direct about it: "if a connection error occurs, you can safely repeat the request without risk of creating a second object".

The header sticks for every attempt, so setting it once does the job. Faked a 500, 500, 200 sequence against the example with the key added:

```
attempts: 3
  #1 status=500 Idempotency-Key=idem-abc-123
  #2 status=500 Idempotency-Key=idem-abc-123
  #3 status=200 Idempotency-Key=idem-abc-123
```

Went with `withHeaders()` over `withHeader()` since that's the one in the 11.x docs.

Item 3 from #895.

### DIFF
--- a/.ai/laravel/skill/laravel-best-practices/rules/http-client.md
+++ b/.ai/laravel/skill/laravel-best-practices/rules/http-client.md
@@ -46,8 +46,11 @@ Correct:
 ```php
 $response = Http::retry([100, 500, 1000])
     ->timeout(10)
+    ->withHeaders(['Idempotency-Key' => $idempotencyKey])
     ->post('https://api.stripe.com/v1/charges', $data);
 ```
+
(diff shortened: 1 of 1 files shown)


pushpak1300:
Dropped the idempotency key here. The tip in this section is about retrying with backoff, and that stands on its own — idempotency is a separate concern that doesn't belong in the retry rule.

Also swapped `api.stripe.com` for `api.example.com`. We don't want to generalize guidance off one specific vendor's API, and the rest of the file already uses `api.example.com`.

Issue #930, opened 2026-08-10
Blank `executable_paths` env vars are treated as real values, silently breaking MCP config and guidelines


sulimanbenhalim:
If you add one of the `executable_paths` env vars to `.env` but leave it blank, Boost treats the empty string as a real path. Two things break and neither of them tells you.

## what happens

`BOOST_PHP_EXECUTABLE_PATH=` gives you `''`, not `null`. Every consumer guards with either `??` or `!== null`, and neither catches `''`.

**mcp config.** The `command` key disappears from the written `.mcp.json`, because `FileWriter` filters empty values out:

```json
// unset
{"mcpServers":{"laravel-boost":{"command":"php","args":["artisan","boost:mcp"]}}}

// BOOST_PHP_EXECUTABLE_PATH=
{"mcpServers":{"laravel-boost":{"args":["artisan","boost:mcp"]}}}
```

`McpWriter::write()` still returns `SUCCESS`, so install reports success and the server just never starts.

**guidelines.** `GuidelineAssist::artisan()` gives back `" artisan"`, and that goes straight into the composed guidelines. in a real app install it turned all 18 occurrences of `php artisan` into ` artisan`:

> Run Artisan commands directly via the command line (e.g., ` artisan route:list`). Use ` artisan list` to discover available commands.

18 blade files call `artisanCommand()`/`artisan()`.

## why

`env()` hands back `''` for a blank value, and for a couple of other spellings too:

| `.env` | `env()` returns | `!== null` |
|---|---|---|
| `KEY=` | `''` | passes |
| `KEY= ` (trailing space) | `''` | passes |
| `KEY=empty` | `''` | passes |
| `KEY=null` | `null` | falls through |
| `KEY=false` | `false` (bool) | passes |

sites:

- `GuidelineAssist.php:125` npm, `:145` composer, `:160` vendor_bin, `:175` php, all `!== null`
- `Agent.php:35` php, `ToolExecutor.php:128` php, `Sail.php:40` sail, all `??`

`Codex.php:80` already filters `''` explicitly (`! in_array($value, [[], null, ''], true)`), so blank is already treated as a real case in one spot. The rest just isn't consistent with it.

Config is registered with `mergeConfigFrom`, so this hits you whether or not you published `config/boost.php`.

## repro

```php
config(['boost.executable_paths.php' => '']);   // what BOOST_PHP_EXECUTABLE_PATH= gives you

[...]

PR #939, opened 2026-08-16
Publish the configuration keys Boost already reads

sulimanbenhalim:
`config/boost.php` publishes 12 keys. The code reads 60.

Some of the gaps are things people would actually want. `mcp.tools.exclude` drops a bundled MCP tool, `mcp.tools.include` registers tool classes of your own, `tinker_tool_enabled` switches on the Tinker tool that ships off, and `purpose` gets prepended to every guideline file Boost generates. Right now you only find any of them by reading src.

So this publishes the eight user-facing ones, same shape as #910:

| key | read at |
|---|---|
| `purpose` | `foundation.blade.php:15` |
| `github.token` | `GitHubSkillProvider.php:283` |
| `enforce_tests` | `InstallCommand.php:200` |
| `mcp.tool_timeout` | `ToolExecutor.php:70` |
| `mcp.tools.include` | `ToolRegistry.php:42` |
| `mcp.tools.exclude` | `ToolRegistry.php:28` |
| `tinker_tool_enabled` | `Tinker.php:29` |

`executable_paths.sail` is the eighth and the odd one out, five of its six siblings were already published.

Every default is the value its call site already falls back to, so publishing this changes nothing until someone edits it.

I left `enforce_tests` and `mcp.tool_timeout` as plain values rather than `env()`, because a blank line in .env breaks both. `BOOST_ENFORCE_TESTS=` gives you `''`, which sails past the `!== null` check and quietly turns enforcement off. Blank timeout is worse, tried it in a real app first:

```
BOOST_MCP_TOOL_TIMEOUT=
getTimeout() : 1 seconds
```

`(int) ''` is 0 and `max(1, 0)` is 1, so every tool would give up after a second. Same class as #930.

Skipped `hosted.api_url` (internal) and `browser_logs` (reads like a legacy alias sitting next to `browser_logs_watcher`, probably wants removing rather than publishing). Left the 38 `agents.*` path keys alone too, happy to do those separately if you want them public.

One caveat worth knowing: `mergeConfigFrom` is a shallow merge, so anyone with an already-published config keeps their copy and won't see these until they republish.

### DIFF
--- a/config/boost.php
+++ b/config/boost.php
@@ -19,6 +19,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Application Purpose
+    |--------------------------------------------------------------------------
+    |
+    | A short description of what your application does. When set, Boost adds
+    | it to the top of the guidelines it generates so agents know what they
+    | are working on before they read any of your application's code.
+    |
+    */
+
+    'purpose' => env('BOOST_PURPOSE'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Boost Project Rules
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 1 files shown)


pushpak1300:
all these are experimental stuff which we don't want to document.

PR #950, opened 2026-08-18
Correct the Wayfinder query parameter return value

sulimanbenhalim:
The query parameter example says `show()` returns a URL string. It returns the route object.

Generated the route with `wayfinder:generate` and ran it:

```
show(1, {query})     -> {"url":"/posts/1?page=1","method":"get"}  typeof: object
show.url(1, {query}) -> "/posts/1?page=1"                         typeof: string
interpolated into an href: [object Object]
```

Only `.url()` gives you the string. The line right above already shows `show(1)` returning an object, so the snippet contradicts itself too.


pushpak1300:
Thanks.

PR #952, opened 2026-08-18
Point the Inertia Form reset props at the version that added them

sulimanbenhalim:
The `<Form>` reset props note tells people to upgrade to v2.2.0. Those props landed in v2.1.2.

From the v2.1.2 release notes: "Add `resetOnError`, `resetOnSuccess`, `setDefaultsOnSuccess` to Form component" (inertiajs/inertia#2514). v2.2.0 was nested prop merging and `<InfiniteScroll>`.

Boost already has the right number in code, `hasFormComponentResets()` checks `gte('2.1.2')`, so the note was pointing further than it needs to. Same line in the vue, react and svelte skills.


pushpak1300:
Thanks for your pull request!

PR #953, opened 2026-08-18
Invert "non-interactive" logic in `boost:install` and `boost:update`

### LINKED ISSUE #948: Guide agents not to pass `-n` to Boost install/update commands
When an agent installs or updates boost, it will often pass the `-n` flag (for non-interactive). This will skip prompting confirmation of adding third-party guidelines, and the agent will happily continue building with none of them compiled (never notifying the user).

It happens because of this guard:

```php
if (! $this->input->isInteractive()) {
    /* early return */
}
```

This guard is obviously intentional and worth having, but ideally having an agent scaffold an app would compile the necessary third-party guidelines. 

This is Claude's take on possible solutions, I'd be happy to contribute one of them if you pick one:

**1. Flip the non-interactive default to "all discovered".** Discovery only finds packages already in composer.json — the user demonstrably uses them, so compiling their guidelines is the sane default; the only cost is tokens. Add --no-third-party (or --packages=none) to opt out. This is exactly the shape of your 38df697 commit: make the useful behaviour the default, keep an opt-out flag, break nothing.

**2. Add an explicit --packages= option.** --packages=livewire,pest or --packages=all|none, with non-interactive defaulting to all. Gives agents and CI a deterministic, testable knob, and it composes with #1 rather than competing.

**3. Print what was decided instead of deciding silently.** Whatever the default, non-interactive runs should emit Included 3 third-party guideline sets (livewire, pest, filament). Run php artisan boost:update to change. Agents read stdout — that's the notification channel that's missing today, and it's independently valuable even if you reject #1.

**4. Same fix in UpdateCommand.** Its early return at line 62 means an agent's boost:update -n also silently ignores newly-installed packages forever. If install defaults to "all", update should too, so a later run recovers.

paulinevos:
Fixes #948 

Currently, when running `boost:install` or `boost:update` in non-interactive mode, it skips compiling (discovered) third-party guidelines altogether, as discovery prompts the user to select which guidelines to include. And an agent will often run these commands interactively, resulting in an agent-scaffolded app never including third-party guidelines (and never communicating this to the user).

But we can logically assume that if a user has installed third-party packages, they will want to use them according to third-party guidelines. This PR therefore inverts the non-interactive logic, and adds an explicit `--packages` option and a `--no-third-party` flag if third-party packages should be explicitly added per package or skipped completely.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -50,7 +50,6 @@ class InstallCommand extends Command
         {--guidelines : Install AI guidelines}
         {--skills : Install agent skills}
         {--mcp : Install MCP server configuration}
-        {--no-third-party : Skip guidelines & skills shipped by third-party packages}
         {--packages= : Third-party packages to include, comma separated, or "all" or "none"}';
 
     /** @var Collection<int, Agent> */
@@ -271,10 +270,6 @@ class InstallCommand extends Command
             return $this->requestedThirdPartyPackages($packages);
         }
 
-        if ($this->option('no-third-party')) {
-            return collect();
-        }
-
         if ($packages->isEmpty()) {
             return collect();
         }
@@ -284,7 +279,9 @@ class InstallCommand extends Command
             ->values();
 
         if (! $this->input->isInteractive()) {
-            return $this->thirdPartyPackagesWithoutPrompt($packages, $defaults);
+            $this->reportSkippedThirdPartyPackages($packages->keys()->diff($defaults));
+
+            return $defaults;
         }
 
         return collect(multiselect(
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks for digging into this.

My main concern is making `-n` include third-party content by default. Since agents commonly run with `-n`, that would effectively bypass the existing trust gate for most installs.

We already treat Laravel and a small set of closely related packages as trusted. I think we should keep that scope narrow for now.

A follow-up could add `spatie/*` and other popular packages we are comfortable trusting, but I would rather not open that up in this PR. It could quickly lead to lots of allowlist requests that we then have to review and maintain in Boost.

paulinevos:
Hey @pushpak1300, sorry it took me a minute to reply! Was at LaravelLive DK. 

I see the concern here, although I would also argue that if the user installs a versioned dependency, and a skill is versioned under that dependency, it would mean that they trust that dependency (including the skill) 🤔 But I'll concede it's kind of a weird thing... I discussed it with someone at LaravelLive and we got onto the topic of, should a new skill constitute a major change under semver? These are all questions nobody has an answer to yet.

It does make me wonder if there's a different solution to the issue that you'd be happy with. I'll close this PR and leave the issue itself open.

PR #954, opened 2026-08-18
Ignore a repository-root SKILL.md so add-skill cannot delete the whole skills directory

sulimanbenhalim:
Running `boost:add-skill` against a repo that keeps its `SKILL.md` at the root deletes everything in `.ai/skills` and installs nothing. That layout is the normal one for a single-skill repo, so this is reachable by just pointing the command at one.

The chain:

1. `discoverSkills()` accepts any blob named `SKILL.md`. For a root-level one, `dirname()` is `.`, so the skill is named `.`.
2. `skillTargetPath()` builds `.ai/skills/.`, which is the skills directory itself.
3. `addSkills()` sees `is_dir()` is true and calls `File::deleteDirectory('.ai/skills/.')` before downloading. That returns `true` and takes every installed skill with it.
4. `downloadSkill()` then bails anyway, because `extractSkillFilesFromTree()` looks for the prefix `./` and finds nothing.

So you end up with an empty `.ai/skills` and "Some skills failed to install", which does not point anywhere near what actually happened.

Ran it end to end against the real command:

```
BEFORE .ai/skills: ["another-skill","my-important-skill"]
AFTER  .ai/skills: []
```

A skill takes its name from its directory, so a root `SKILL.md` has no name to take. Skipping it at discovery removes the whole path. There is already a test called "ignores files at root level", it just uses README/LICENSE and never a root `SKILL.md`.

Three tests added: root `SKILL.md` is ignored, nested skills still work when a root one is present, and an end to end one asserting an installed skill survives.

### DIFF
--- a/src/Skills/Remote/GitHubSkillProvider.php
+++ b/src/Skills/Remote/GitHubSkillProvider.php
@@ -38,7 +38,10 @@ class GitHubSkillProvider
         $basePath = $this->repository->path;
 
         $skillMarkers = collect($tree['tree'])
-            ->filter(fn (array $item): bool => $item['type'] === 'blob' && basename((string) $item['path']) === 'SKILL.md');
+            ->filter(fn (array $item): bool => $item['type'] === 'blob'
+                && basename((string) $item['path']) === 'SKILL.md'
+                // A skill is named after its directory, so a repository-root SKILL.md has no name to take.
+                && dirname((string) $item['path']) !== '.');
 
         if ($basePath !== '') {
             $prefix = $basePath.'/';
--- a/src/Console/AddSkillCommand.php
+++ b/src/Console/AddSkillCommand.php
@@ -100,7 +100,7 @@ class AddSkillCommand extends Command
         }
 
         if ($this->availableSkills->isEmpty()) {
-            $this->error('No valid skills are found in the repository. Remote skills must provide a SKILL.md; SKILL.blade.php is not supported.');
+            $this->error('No valid skills are found in the repository. Each skill must live in its own directory containing a SKILL.md — a SKILL.md at the repository root is not a skill. SKILL.blade.php is not supported.');
 
             return false;
         }
--- a/tests/Feature/Console/AddSkillCommandTest.php
+++ b/tests/Feature/Console/AddSkillCommandTest.php
@@ -508,7 +508,6 @@ it('displays error when rate limit is exceeded', function (): void {
 });
 
 it('does not wipe installed skills when the repository has a root level SKILL.md', function (): void {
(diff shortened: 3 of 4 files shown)


src/Skills/Remote/GitHubSkillProvider.php
```diff
@@ -38,7 +38,10 @@ public function discoverSkills(): Collection
         $basePath = $this->repository->path;
 
         $skillMarkers = collect($tree['tree'])
-            ->filter(fn (array $item): bool => $item['type'] === 'blob' && basename((string) $item['path']) === 'SKILL.md');
+            ->filter(fn (array $item): bool => $item['type'] === 'blob'
+                && basename((string) $item['path']) === 'SKILL.md'
+                // A skill is named after its directory, so a repository-root SKILL.md has no name to take.
+                && dirname((string) $item['path']) !== '.');
```
pushpak1300, line 44: The user's repo is a skill, and now we tell them it isn't which imo is incorrect behaviour. 

also did you identified any real repo where this is actually issues ? 

can you share example ?

sulimanbenhalim: Yeah there are real ones and `ProjectOpenSea/opensea-skill` is the one that bugs me because it has a root `SKILL.md` plus 5 nested skills so `--all --force` for the 5 real ones drags the root marker in too. Same shape in `cdxgen/cdxgen` and code search shows around 38k root-level `SKILL.md` files.

Thing is we can't install a root `SKILL.md` today anyway because `discoverSkills()` names it `.` and `downloadSkill()` then always fails since `extractSkillFilesFromTree()` builds the prefix `./` which matches nothing in a git tree. So the only part that actually runs is the delete where `.ai/skills/.` resolves to `.ai/skills` and `addSkills()` wipes it first and nothing tells you your other skills are gone.

Agreed repo-as-skill should work though and it's a bit more than a rename since `extractSkillFilesFromTree()` and `getRelativePath()` both assume a `basePath/` prefix. Happy to extend this PR that way or keep it as the safety fix, your call.

sulimanbenhalim: A Laravel one is `agarzon/moonshine-v4-skills`, a MoonShine admin panel with a root `SKILL.md` plus 8 nested. Seeded `.ai/skills` with two of my own then ran `--all --force`:

```
BEFORE: company-standards, my-precious-skill
AFTER : moonshine-fields-v4

Skills installed:
 ┌─────────────────────┐
 │ moonshine-fields-v4 │
 └─────────────────────┘
                       Enjoy the boost 🚀
Some skills failed to install:
 ┌───┐
 │ . │
 └───┘
```

Nothing in the output mentions the delete.

PR #955, opened 2026-08-18
Reject a skill named dot so it cannot resolve to the skills directory itself

sulimanbenhalim:
`isValidSkillName()` blocks `..`, `/`, `\` and blank names, but not `.`. A skill named `.` builds the target path `.claude/skills/.`, which is the skills directory itself, so `remove('.')` deletes every skill the agent has and then returns `false` because the final `rmdir` fails.

```
BEFORE: ["also-keep","keep-me"]
remove('.') returned: false
AFTER:  []
```

It is reachable through `boost.json`. `Config::getSkills()` reads the `skills` array as is, `sync()` diffs it against what was just discovered, and anything left over goes to `removeStale()`. So a `"."` entry in that file wipes `.claude/skills`, `.cursor/skills` and `.agents/skills` on the next `boost:install` or `boost:update`. `boost.json` is committed and hand editable, so it does not take much to get there.

Added `.` to the rejected list. Two tests, one on `remove()` and one through `removeStale()` to cover the `boost.json` route. Both assert the directory contents survive, not just that the call returns `false`, since the old code also returned `false` after deleting.

### DIFF
--- a/src/Install/SkillWriter.php
+++ b/src/Install/SkillWriter.php
@@ -31,7 +31,7 @@ class SkillWriter
 
     public function write(Skill $skill): int
     {
-        if (! $this->isValidSkillName($skill->name)) {
+        if (! self::isValidSkillName($skill->name)) {
             throw new RuntimeException("Invalid skill name: {$skill->name}");
         }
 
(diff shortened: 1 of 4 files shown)


pushpak1300:
https://github.com/laravel/boost/pull/954#discussion_r3809609307 ? any example of this type of repos ?

sulimanbenhalim:
Different trigger here, it's a `boost.json` entry not a repo, so the examples are over in #954.

Can't show a path that gets `.` in there today though, `glob('*', GLOB_ONLYDIR)` never returns dot entries so it'd take a hand-edited file. Only argument for keeping it is that `isValidSkillName()` already blocks `..` and `/` and `\` for the same reason and `.` slips through, then `remove()` deletes `.ai/skills/.` and takes every skill for that agent. Fine to close if that's too thin.

pushpak1300:
Closing in favour of #960 — the changes here are ported into that branch.

#960 touches the same few lines of `GitHubSkillProvider::discoverSkills()`, so keeping both open guaranteed a conflict for whichever landed second. Ported as-is: the `.` rejection (plus `\0` and dot/whitespace-only names), `isValidSkillName()` becoming public static, the marker filter routing remote skill names through it, and `writeAll()`/`sync()` letting one bad name fail without blocking the skills beside it. Your two tests came across too.

Thanks for catching it — the `remove(".")` case was a genuine "deletes every installed skill" bug, and gating remote discovery on the same predicate closes the same class of hole for skills fetched from a repo.

PR #958, opened 2026-08-20
Quote the MySQL table type as a string literal so schema reads work under ANSI_QUOTES

sulimanbenhalim:
`getTables()` compares `TABLE_TYPE` against `"BASE TABLE"`. With `ANSI_QUOTES` in `sql_mode` those double quotes are an identifier rather than a string, so the query throws and the `catch (Exception)` right below it turns that into an empty table list.

MySQL 9.2.0, same two table database:

```
ANSI_QUOTES on,  "BASE TABLE"  ->  ERROR 1054 Unknown column 'BASE TABLE' in 'where clause'
ANSI_QUOTES on,  'BASE TABLE'  ->  posts, users
```

driver level before and after:

| sql_mode | `getTables()` today | with the fix |
|---|---|---|
| `ANSI_QUOTES` | 0 tables | 2 tables |
| default | 2 tables | 2 tables |

`getTables()` is the only source of the table list for both summary and full mode so the whole schema comes back empty. `ANSI` resolves to `REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,ONLY_FULL_GROUP_BY,ANSI` so that spelling hits it too. The PostgreSQL driver next to it already uses single quotes.

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema/MySQLSchemaDriver.php
+++ b/src/Mcp/Tools/DatabaseSchema/MySQLSchemaDriver.php
@@ -75,13 +75,13 @@ class MySQLSchemaDriver extends DatabaseSchemaDriver
     public function getTables(): array
     {
         try {
-            return DB::connection($this->connection)->select('
+            return DB::connection($this->connection)->select("
                 SELECT TABLE_NAME as name
                 FROM information_schema.TABLES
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks

PR #959, opened 2026-08-20
Point the last-error tool description at the correct browser-logs tool name

sulimanbenhalim:
The `last-error` description tells the agent to use the `browser-log` tool. The registered name is `browser-logs`.

```
BrowserLogs registered name : browser-logs
LastError description       : ... Use browser-log tool for browser errors.
```

`browser-log` appears nowhere else in `src/` so there's nothing under that name to call. `laravel/mcp` derives the name from `Str::kebab(class_basename($this))`, so the test asserts the description contains `(new BrowserLogs)->name()` instead of a hardcoded string.

### DIFF
--- a/src/Mcp/Tools/LastError.php
+++ b/src/Mcp/Tools/LastError.php
@@ -46,7 +46,7 @@ class LastError extends Tool
     /**
      * The tool's description.
      */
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks

PR #961, opened 2026-08-21
Leave fenced code blocks alone when spacing markdown headings

sulimanbenhalim:
`MarkdownFormatter::format()` spaces out anything that looks like a heading, but it doesn't know about code fences. So a `# comment` line inside a bash block gets blank lines shoved around it.

This ships in the box today. Real diff for `.ai/laravel/skill/laravel-best-practices/rules/config.md`:

````diff
 Incorrect:
 ```bash
+
 # .env committed to repo or shared in Slack
+
 STRIPE_SECRET=sk_live_abc123
````

The Livewire 4 skill is worse, `.ai/livewire/4/skill/livewire-development/SKILL.blade.php` has 9 comment lines in one bash block and every one gets split. User content goes through it too, `SkillWriter::copyFile()` and `RuleComposer::composeManaged()` both format everything, so a recorded rule with a code sample gets mangled the same way.

Fix masks fences before the heading pass and restores them after, same trick `RendersBladeGuidelines::markScopedBlocks()` already uses. Also anchored the second heading regex to line start so `Use # for comments` in normal prose isn't read as a heading, the first one already is anchored.

Side effect worth calling out: blank lines inside a fence now survive the `\n{3,}` collapse.

8 tests, 4 for the bug and 4 pinning what already worked (headings still spaced, blank lines still collapsed outside fences, CRLF still normalized). Reverting the src change fails exactly the first 4.

### CODE REVIEW

**Comment by sulimanbenhalim** [comment]
ran this end to end on a fresh laravel 13 app with livewire 4.4.1 installed, `boost:install --skills -n`, before and after.

what lands in `.claude/skills/livewire-development/SKILL.md` on main:

```bash

# Single-file component (SFC - default in v4)

# Creates: resources/views/components/⚡create-post.blade.php

php artisan make:livewire create-post

# Page component (SFC - Full Page in v4)
```

and on this branch:

```bash
# Single-file component (SFC - default in v4)
# Creates: resources/views/components/⚡create-post.blade.php
php artisan make:livewire create-post

# Page component (SFC - Full Page in v4)
```

2 of the 4 skills that install by default are affected, 13 stray blank lines total across `livewire-development/SKILL.md` and `laravel-best-practices/rules/config.md`. diffed the whole installed tree, every non-blank line is identical before and after, so nothing else moves.

PR #962, opened 2026-08-21
Re-key the search-docs query list after filtering

sulimanbenhalim:
`SearchDocs` drops empty and `*` queries with `array_filter`, which keeps the original keys. so if the dropped one isnt last, `queries` stops being a json array and goes out as an object.

real body for `queries => ['*', 'middleware', '', 'routing']`:

```json
{"queries":{"1":"middleware","3":"routing"},"packages":[{"name":"laravel/framework","version":"11.x"}],"token_limit":3000,"format":"markdown"}
```

`packages` right next to it is fine because it already gets `->values()`, so the list shape is clearly what was intended here too.

worth saying i cant see what the docs api does with an object instead of an array, that ones on your side. but the schema declares `queries` as an array of strings and thats not what it sends.

fix is `array_values()` around the existing filter. test asserts both the decoded value and the literal `"queries":["middleware","routing"]` in the body, and it fails without the change.

### DIFF
--- a/src/Mcp/Tools/SearchDocs.php
+++ b/src/Mcp/Tools/SearchDocs.php
@@ -68,7 +68,6 @@ class SearchDocs extends Tool
             return $rawQueries;
         }
 
-        // Re-key so a dropped query cannot turn the list into a JSON object.
         $queries = array_values(array_filter(
             array_map(trim(...), $rawQueries),
             fn (string $query): bool => $query !== '' && $query !== '*'
--- a/src/Mcp/Tools/SearchDocs.php
+++ b/src/Mcp/Tools/SearchDocs.php
@@ -68,10 +68,11 @@ class SearchDocs extends Tool
             return $rawQueries;
         }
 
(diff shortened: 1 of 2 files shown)


sulimanbenhalim:
tested this against the real api on a fresh laravel 13 app, so i can answer my own question from the description. the object form just fails.

`queries: ["*", "middleware"]` on main:

```
HTTP 200
# Something went wrong
<error>Failed to search, please try again</error>
```

same input on this branch:

```
HTTP 200
# Search Results

## Middleware Registering Middleware
Middleware | laravel/framework@13.x | Source: https://github.com/laravel/docs/blob/13.x/middleware.md
```

`queries: ["middleware"]` on its own works either way, so its specifically the re-keying that kills it.

one extra thing worth knowing: it comes back as HTTP 200 with the error in the body, so `$response->successful()` is true and the tool hands "Something went wrong" to the agent as if it were documentation. no error surfaces anywhere.

PR #966, opened 2026-08-24
Stop detecting Antigravity from the shared .agents directory

sulimanbenhalim:
`.agents` isnt an antigravity-specific signal. amp, codex, opencode and zed all install their skills into `.agents/skills`, so boost itself creates that directory, and after that antigravity shows up as a project-installed agent on every rerun.

concrete case: fresh app, `boost:install`, pick codex + skills. that writes `.agents/skills/...`. rerun `boost:install -n` and antigravity is now in the agent selection automatically, project-detected agents become the non-interactive defaults (`InstallCommand::selectAgents`), so boost configures an IDE that was never installed or selected. interactive reruns get it pre-checked in the multiselect too.

same bug class as #831/#832, which dropped `AGENTS.md` from codex/opencode project detection because other agents write that file. the fix here is the same shape: keep `.agents/mcp_config.json` as the antigravity signal (nothing else writes that path) and drop the bare `.agents` path check.

updated the pinning test and added regression tests both ways: `.agents/skills` alone should not detect, `.agents/mcp_config.json` should.

PR #967, opened 2026-08-24
Read the standard mcp_config_path config key for Junie

sulimanbenhalim:
junie is the only agent reading `boost.agents.junie.mcp_path`. the other 11 mcp-capable agents all read `mcp_config_path`, and junie itself already uses the standard names for `guidelines_path` and `skills_path`.

how it happened: #626 gave junie a configurable `mcp_path`, then #729 standardised on `mcp_config_path` across all agents but skipped junie since it already had an override under the old name. `mcp_path` never made it into any docs or config file, that line is its only occurrence in the repo.

the failure is silent and asymmetric: in the monorepo setup #729 was built for, setting `boost.agents.junie.mcp_config_path` works for every agent except junie, which quietly keeps writing to the default `.junie/mcp/mcp.json` and the IDE never sees the server.

switched the key and added the missing `mcpConfigPath` coverage to JunieTest.

### DIFF
--- a/src/Install/Agents/Junie.php
+++ b/src/Install/Agents/Junie.php
@@ -59,7 +59,7 @@ class Junie extends Agent implements SupportsGuidelines, SupportsMcp, SupportsSk
 
     public function mcpConfigPath(): string
     {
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #968, opened 2026-08-24
Map check constraints through TABLE_CONSTRAINTS so they work on MySQL

sulimanbenhalim:
`database-schema` reports zero check constraints for every table on MySQL, always. the query filters `information_schema.CHECK_CONSTRAINTS` by `TABLE_NAME`, but that column only exists on mariadb. mysql 8.0's `CHECK_CONSTRAINTS` has exactly four columns (`CONSTRAINT_CATALOG`, `CONSTRAINT_SCHEMA`, `CONSTRAINT_NAME`, `CHECK_CLAUSE`), the table mapping lives in `TABLE_CONSTRAINTS`. the `catch (Exception)` in `getCheckConstraints` swallows the error so nothing ever surfaces, and on mariadb the same code works, which is probably why nobody noticed.

reproduced on real servers. mysql 8.0.46, `CREATE TABLE orders (total INT, CONSTRAINT chk_total CHECK (total >= 0))`, then the query from main:

```
ERROR 1054 (42S22): Unknown column 'TABLE_NAME' in 'where clause'
```

so the tool hands the agent `"check_constraints": []` for every table and it happily writes inserts that violate them.

the fix joins `TABLE_CONSTRAINTS` on `(CONSTRAINT_SCHEMA, CONSTRAINT_NAME)` and filters `tc.TABLE_NAME` + `tc.CONSTRAINT_TYPE = 'CHECK'`. same query verified on both engines:

mysql 8.0.46:

```
CONSTRAINT_NAME	CHECK_CLAUSE
chk_total	(`total` >= 0)
```

mariadb 11.8.8:

```
CONSTRAINT_NAME	CHECK_CLAUSE
chk_total	`total` >= 0
```

docs reference: https://dev.mysql.com/doc/refman/8.0/en/information-schema-check-constraints-table.html

separate from #958, which only touched `getTables()`. added a unit test in the same capture-the-sql style pinning the join and bindings.

### DIFF
--- a/src/Mcp/Tools/DatabaseSchema/MySQLSchemaDriver.php
+++ b/src/Mcp/Tools/DatabaseSchema/MySQLSchemaDriver.php
@@ -56,12 +56,18 @@ class MySQLSchemaDriver extends DatabaseSchemaDriver
     public function getCheckConstraints(string $table): array
     {
         try {
+            // MySQL's CHECK_CONSTRAINTS table has no TABLE_NAME column (MariaDB's does),
+            // so the table mapping must come from TABLE_CONSTRAINTS on both engines.
             return DB::connection($this->connection)->select('
-                SELECT CONSTRAINT_NAME, CHECK_CLAUSE
-                FROM information_schema.CHECK_CONSTRAINTS
-                WHERE CONSTRAINT_SCHEMA = DATABASE()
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #970, opened 2026-08-24
Report the database driver as database_engine in application-info

sulimanbenhalim:
`application-info` returns `config('database.default')` as `database_engine`, but thats the connection *name*, not the driver. it only looks right on stock installs because the default connection names happen to match the driver names.

any app with `DB_CONNECTION=tenant` (or `landlord`, `primary`, `read`, whatever) where the connection's driver is `pgsql` gets `"database_engine": "tenant"` in the tool the description tells agents to call on every new chat. no usable dialect signal. boost's own `DatabaseSchema` already does this correctly via `getDriverName()` (`DatabaseSchema.php:104`), and the test suite itself runs on a connection literally named `testing`, so on main the tool reports `"database_engine": "testing"` in this repo's own tests.

switched to `DB::connection()->getDriverName()`, pinned the existing expectations to `'sqlite'` instead of mirroring the implementation, and added a regression test with a custom-named default connection that fails on main.

### DIFF
--- a/src/Mcp/Tools/ApplicationInfo.php
+++ b/src/Mcp/Tools/ApplicationInfo.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Laravel\Boost\Mcp\Tools;
 
+use Illuminate\Support\Facades\DB;
 use Laravel\Boost\Support\PackageRegistry;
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #971, opened 2026-08-24
Keep the separating comma out of trailing comments when injecting MCP servers

sulimanbenhalim:
if the last thing inside `mcpServers` before the closing brace is a `//` comment, `boost:install` corrupts the whole config file. `findCommaInsertionPoint` walks backwards from the closing brace, but its comment check only fires when it lands exactly on a `/`, and walking backwards through a comment you land on the comment's last character, so the comma gets inserted mid-comment.

repro through the real writer, input `.cursor/mcp.json`:

```jsonc
{
  "mcpServers": {
    "context7": {
      "command": "npx"
    } // docs lookup
  }
}
```

what main writes:

```jsonc
    } // docs lookup,
    "laravel-boost": {
```

the comma is commented out, the file no longer parses, and every mcp server the user had stops loading. rerunning `boost:install` doesnt repair it either, `serverExistsInContent` sees `"laravel-boost":` in the broken file and save() returns true without touching it.

`needsCommaBeforeClosingBrace` handles the same case correctly (it strips comments before checking), so the two helpers just disagreed about where a comment ends.

fix masks comments before the backwards scan, reusing the string-aware pattern `hasUnquotedComments` already uses so a `//` inside a `"url": "https://..."` value doesnt get treated as one. when a trailing comment sits between the last entry and the brace, the new server block now goes after the comment line, so the output is:

```jsonc
    }, // docs lookup

    "laravel-boost": {
```

comment-free files produce byte-identical output to main. added a regression test that fails on main with the exact repro above, all 68 existing FileWriter tests still pass.

### DIFF
--- a/src/Install/Mcp/FileWriter.php
+++ b/src/Install/Mcp/FileWriter.php
@@ -150,8 +150,15 @@ class FileWriter
         $commaPosition = $this->findCommaInsertionPoint($content, $openBracePos, $closeBracePos);
 
         if ($commaPosition !== -1) {
-            $newContent = substr_replace($content, ',', $commaPosition, 0);
-            $newContent = substr_replace($newContent, $serversJson, $commaPosition + 1, 0);
+            if (trim(substr($content, $commaPosition, $closeBracePos - $commaPosition)) === '') {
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #975, opened 2026-08-24
Leave fenced code blocks alone when normalizing blank lines in guideline files

sulimanbenhalim:
`GuidelineWriter::write` runs `preg_replace("/\n{3,}/", "\n\n")` over the whole target file, not just the `<laravel-boost-guidelines>` block, and its not fence-aware. so if your `AGENTS.md`/`CLAUDE.md` has a fenced code sample with a deliberate double blank line (pep 8 mandates two blank lines between top level functions, for example), every `boost:install` or `boost:update` silently rewrites your own code sample and leaves a spurious diff in your repo.

same bug class as #961, which fixed the fence-unaware heading spacing in `MarkdownFormatter`, and the fix uses the same fence masking mechanism: mask ``` / ~~~ blocks with placeholders, collapse blank lines, restore. normalization outside fences behaves exactly as before.

regression test writes guidelines into a file whose existing content has a python fence with two blank lines, and asserts the fence survives byte for byte. fails on main.

### DIFF
--- a/src/Install/GuidelineWriter.php
+++ b/src/Install/GuidelineWriter.php
@@ -76,8 +76,21 @@ class GuidelineWriter
                 $newContent = $frontMatter.$existingContent.$separatingNewlines.$replacement;
             }
 
-            // Normalize multiple blank lines to single blank lines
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #977, opened 2026-08-24
Keep the separating comma out of trailing comments when injecting MCP servers

sulimanbenhalim:
if the last thing inside `mcpServers` before the closing brace is a `//` comment, `boost:install` writes the separating comma into the comment and corrupts the whole file.

repro on a fresh laravel app, `.cursor/mcp.json` has:

    "context7": {
      "command": "npx"
    } // docs lookup

after `php artisan boost:install --mcp -n`:

    } // docs lookup,
    "laravel-boost": {

the comma is commented out, the file no longer parses, every mcp server the user had stops loading, and a rerun doesnt repair it since the broken file already contains `"laravel-boost":`.

fix masks comments before finding the comma position (string-aware, so a `//` inside a `"url"` value doesnt count) and puts the new server block after the comment line. comment-free files produce byte-identical output to main. regression test fails on main, all 68 existing FileWriter tests still pass. previously #971, ci was green there.

### DIFF
--- a/src/Install/Mcp/FileWriter.php
+++ b/src/Install/Mcp/FileWriter.php
@@ -297,9 +297,10 @@ class FileWriter
     protected function findCommaInsertionPoint(string $content, int $openBracePos, int $closeBracePos): int
     {
         // Blank out comments (string-aware, offsets preserved) so the backwards
-        // scan cannot place the comma inside a trailing comment.
+        // scan cannot place the comma inside a trailing comment. Both quote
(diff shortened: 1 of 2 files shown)


sulimanbenhalim:
re-tested and found my own fix broke single-quoted configs, pushed 9fa89cc.

```
before:  'url': 'https:,//example.test/sse' }
after:   'url': 'https://example.test/sse' },
```

my regression, not the original bug. test added.

pushpak1300:
Thanks. feel free to send other chnages.

PR #978, opened 2026-08-24
Consolidate test guidance into one skill for Pest and PHPUnit

pushpak1300:
Boost's own guidelines were pushing agents to over-test. `enforce-tests` said every change must be programmatically tested with no bound, so on a greenfield build every file counted as a change. The `pest-testing` skill then showcased browser, smoke, and visual regression testing with no cost signal and no statement that the plugin was absent. A `laravel new` followed by "build a portfolio" produced a flood of tests plus an unprompted install of the Pest browser plugin and Chrome.

There was also no test design guidance anywhere in Boost, so agents wrote change detector tests: assertions on config arrays, route definitions, `$fillable`, and Blade markup. Those only fail when the declaration changes, so they break on every refactor while verifying nothing.

PHPUnit projects had it worse. They shipped no test skill at all, and `.ai/phpunit/core.blade.php` told them to "cover all happy paths, failure paths, and edge cases".

### What this adds

One `testing-best-practices` skill under `.ai/laravel/`, gated on `laravel/framework` so it reaches Pest and PHPUnit projects alike. It is a thin index over nine rule files, each branching on the installed framework:

```blade
$pest = $assist->hasPackage('pestphp/pest');
```

A Pest project reads `it()` and `test()`, `beforeEach()`, datasets, and `expect()` chains. A PHPUnit project reads `test_` prefixed methods, a private factory method instead of records in `setUp()`, `#[DataProvider]`, and `assertSame()`. Browser guidance gates on the package that actually provides it, `pestphp/pest-plugin-browser` for Pest and `laravel/dusk` for PHPUnit, and otherwise tells the agent to write feature tests rather than install a browser.

### What this removes

`.ai/pest/3/skill/pest-testing` and `.ai/pest/4/skill/pest-testing`. Their sources differed only in the heading digit, one table cell, and the browser plugin blocks, and their `reference/performance.md` files were byte identical. Most of their content restated `.ai/pest/core.blade.php`, which is `@scoped(['tests/**'])` and so is always loaded when a test file is touched, a stronger guarantee than a skill trigger.

`laravel-best-practices/rules/testing.md`, absorbed into the new skill. That skill triggers on controllers, models, and policies rather than tests, so it was never a reliable home for guidance that has to fire on every test write.

Both Pest skills also stopped repeating pestphp.com. They now teach discovery: search by capability, and report that a feature does not exist when the search returns nothing.

### A latent bug this fixes

Skill discovery maps a package major straight onto a directory name, with no fallback:

```php
$this->discoverSkillsFromPath($aiPath, $name, (string) $package->major());
// reads {package}/skill and {package}/{major}/skill
```

Because `.ai/pest/skill/` never existed, every Pest skill lived only under a numbered directory. A probe skill placed at `.ai/pest/4/skill/zz-probe` is discovered for Pest 4.1.0 and not discovered for 5.0.0, so a Pest 5 project would have received no Pest skill at all, with no error. Guidelines never had this problem, because `{package}/core` always loads and versioned files are additive.

The new skill gates on the package rather than the major, and the tests pin Pest 3, 4, 5, and 6.

### Pest 5

`GuidelineAssist::hasPackage()` now takes an optional constraint, matching `DiscoverPackagePaths::usesPackage()`, which already had that shape:

```php
$pest5 = $assist->hasPackage('pestphp/pest', '>=5.0');
```

That gates Test Impact Analysis (`pest --parallel --tia`) and time balanced sharding, since recommending a command that does not exist is worse than saying nothing. Search terms stay ungated, because a search that returns nothing is already handled.

Pest 5 ships its own `pest-testing` skill in `resources/boost/skills`, which Boost picks up through `resolveFirstPartyBoostPath()`, so Pest 5 users lose nothing here. A follow up PR will remove it there.

### Compatibility

`hasPackage()` keeps working with a single argument, so every existing call site is unaffected. Existing installs clean themselves up, because `SkillWriter::sync()` passes the difference between tracked and current skills to `removeStale()`, so the stale `pest-testing` directory is removed on the next `boost:install` instead of being orphaned.

This also corrects `Http::preventingStrayRequests()`, which returns a bool, to `Http::preventStrayRequests()`. Reported by @jackwh on #769.

### Tests

`tests/Feature/Install/TestingSkillsTest.php` renders every rule file in both framework configurations and asserts that neither leaks the other's syntax, checks the browser gate both ways for both packages, pins the Pest 5 gate to Pest 5, verifies every rule index pointer resolves to a file on disk and that no rule file is unreferenced, and confirms the skill ships to a project with no Pest.

Carries @nexxai's commits from #769 and supersedes it. Addresses #662.


pushpak1300:
Superseded by #769, which now carries the same commit. Closed to avoid two open PRs with identical content.

PR #983, opened 2026-08-25
Document flexible cache stampede limitation

jonhassall:
Cache::flexible does not have protection against cold and stale cache stampedes.

This change gives agents this knowledge, to make informed coding decisions.


pushpak1300:
Thanks for your pull request!

PR #984, opened 2026-08-26
Detect Livewire navigate requests by header presence since Livewire 3 sends an empty value

sulimanbenhalim:
the navigate guard from #734 never fires on Livewire 3. its navigate requests send X-Livewire-Navigate with an empty value (Livewire 4 sends '1'), so the `=== '1'` check misses them and the browser logger script gets injected into navigate responses anyway, which is what #734 was added to prevent.

fix checks for the header's presence instead of its value.

repro on a real app with artisan serve: a request with the empty header gets browser-logger-active on main and not with this change, plain page loads still get the script, and a '1' header is skipped both ways.

re-file of #969, reviewed and re-tested individually.

### DIFF
--- a/src/Middleware/InjectBoost.php
+++ b/src/Middleware/InjectBoost.php
@@ -36,7 +36,6 @@ class InjectBoost
 
     protected function shouldInject(Request $request, Response $response): bool
     {
-        // Livewire 3 sends the navigate header with an empty value, so only its presence is reliable.
         if ($request->headers->has('x-livewire-navigate')) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/Middleware/InjectBoost.php
```diff
@@ -36,7 +36,8 @@ public function handle(Request $request, Closure $next): Response
 
     protected function shouldInject(Request $request, Response $response): bool
     {
-        if ($request->headers->get('x-livewire-navigate') === '1') {
+        // Livewire 3 sends the navigate header with an empty value, so only its presence is reliable.
```
pushpak1300: can you remove this unnecessary comment ?

pushpak1300:
Thanks.

PR #987, opened 2026-08-28
Collapse Blade escape sequences in boostsnippet bodies since they never reach Blade

sulimanbenhalim:
skills that use @boostsnippet ship Blade escape sequences to users. snippet bodies get pulled out before Blade renders the file and restored raw after, so Blade never collapses @@ or @{{ inside them. the volt skill ships literal @@volt, @@endvolt and @{{ $count }}, and the livewire 4 skill ships @{{ $count }}, which is not valid Volt or Livewire code for the agent reading the skill.

fix collapses the two escapes in the stored body the same way Blade would, @@directive to @directive and @{{ to {{.

repro on a real app with livewire 4.4: boost:install --skills lands .claude/skills/livewire-development/SKILL.md with `Count: @{{ $count }}` on main and `Count: {{ $count }}` with this change.

re-file of #972, reviewed and re-tested individually.

### DIFF
--- a/src/Concerns/RendersBladeGuidelines.php
+++ b/src/Concerns/RendersBladeGuidelines.php
@@ -63,6 +63,9 @@ trait RendersBladeGuidelines
             $lang = empty($matches['lang']) ? 'html' : $matches['lang'];
             $snippetContent = trim($matches['content']);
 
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for your pull request!

PR #996, opened 2026-09-03
Stop normalizing blank lines outside the Boost guidelines block

sulimanbenhalim:
`GuidelineWriter::write` ran `preg_replace("/\n{3,}/", "\n\n")` over the **whole target file**, not just the `<laravel-boost-guidelines>` block. Every byte of your hand-written `AGENTS.md`/`CLAUDE.md` got reformatted on every `boost:install` and `boost:update`, leaving a diff you never asked for. Boost was rewriting a file section it does not own.

The most visible casualty is fenced code: a sample with a deliberate double blank line (PEP 8 mandates two between top-level defs) gets silently squeezed. But the fence case is the symptom, not the bug.

## The fix

Delete the normalization instead of making it fence-aware. The Boost block is already normalized upstream:

```
GuidelineComposer::compose() -> MarkdownFormatter::format() -> /\n{3,}/ -> "\n\n"
                                        |
                                GuidelineWriter::write()   <- second squeeze, pure collateral damage
```

The writer's pass only ever touched content it had no claim to, so it goes away. `MarkdownFormatter` (fixed in #961) remains the single place blank lines are normalized, and it only sees content Boost is assembling.

Also extracts the fence-masking regex that #961 added into `Laravel\Boost\Support\Fences`, since `MarkdownFormatter` and `RendersBladeGuidelines` had copy-pasted it. After this PR there is exactly one fence-unaware normalizer left in `src/`, and it runs inside the mask.

## Behavior change

Files that already accumulated triple newlines from older Boost versions keep them. That is intentional: it is not our file.

Two things Boost still normalizes, both deliberate: `MarkdownFormatter` collapses blank lines in user `.ai/guidelines/*.md` prose outside fences (Boost is assembling that into its own block), and the writer still `rtrim()`s the file before appending its `===` separator.

## Tests

Two regression tests, both failing on main:

- append path (no existing block) - a user file with a Python fence and a prose ``` run survives byte for byte
- replace path (existing block) - the one that runs on every `boost:update`, and the one the old squeeze hit repeatedly

Re-file of #975, reviewed and re-tested individually.

### DIFF
--- a/src/Concerns/RendersBladeGuidelines.php
+++ b/src/Concerns/RendersBladeGuidelines.php
@@ -79,11 +79,9 @@ trait RendersBladeGuidelines
                 '/(?<!@)@scoped\(\s*(?P<paths>\[(?:[\s,]|\'[^\']*\'|"[^"]*")*\])\s*\)/s',
                 fn (array $matches): string => '___SCOPED_START_'.base64_encode((string) json_encode($this->parseScopedPaths($matches['paths']))).'___',
                 $markdown
-            );
+            ) ?? $markdown;
 
-            $marked = $marked === null ? null : preg_replace('/(?<!@)@endscoped/', '___SCOPED_END___', $marked);
-
-            return $marked ?? $markdown;
+            return preg_replace('/(?<!@)@endscoped/', '___SCOPED_END___', $marked) ?? $marked;
         });
     }
 
--- a/src/Concerns/RendersBladeGuidelines.php
+++ b/src/Concerns/RendersBladeGuidelines.php
@@ -6,6 +6,7 @@ namespace Laravel\Boost\Concerns;
 
 use Illuminate\Support\Facades\Blade;
 use Laravel\Boost\Install\GuidelineAssist;
+use Laravel\Boost\Support\Fences;
 use Laravel\Boost\Support\RenderFailures;
 
 trait RendersBladeGuidelines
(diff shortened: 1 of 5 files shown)


pushpak1300:
Reviewed this and pushed two small follow-ups to the branch (0d598a6) — happy to drop them if you'd rather apply your own version.

**Deleting the normalization rather than making it fence-aware is the right call.** The first pass on this branch made `GuidelineWriter` fence-aware, which keeps the writer normalizing content it doesn't own, just more carefully. `3bf4949` removing it outright is better: the block is already normalized upstream by `MarkdownFormatter::format()` via `GuidelineComposer::compose()`, so the writer's pass was only ever collateral damage on user content. Pulling the fence regex into `Support\Fences` is a real dedupe too — I grepped `src/` and there's exactly one fence-unaware normalizer left, and it runs inside the mask.

Two things I changed:

**1. The regression test covered the wrong path.** The existing test exercises the *append* path (no existing `<laravel-boost-guidelines>` block) — that's the first install only. The **replace** path is what runs on every `boost:update` after that, and it's where the squeeze did repeated damage. `it preserves user content after guidelines when replacing` asserts with `toContain` on individual fragments, so it says nothing about blank lines and passes on main either way.

Added `it leaves user content around an existing block untouched when replacing`. Confirmed it fails on main and passes here.

**2. Simplified the null handling in `markScopedBlocks`.** The `=== null ? null :` ternary is noise — `preg_replace` only returns null on a regex error and both patterns are fixed literals:

```php
$marked = preg_replace_callback(
    '/(?<!@)@scoped\(\s*(?P<paths>\[(?:[\s,]|\'[^\']*\'|"[^"]*")*\])\s*\)/s',
    fn (array $matches): string => '___SCOPED_START_'.base64_encode((string) json_encode($this->parseScopedPaths($matches['paths']))).'___',
    $markdown
) ?? $markdown;

return preg_replace('/(?<!@)@endscoped/', '___SCOPED_END___', $marked) ?? $marked;
```

I also retitled the PR and rewrote the description — both still described the abandoned fence-masking approach, so a reviewer would go looking for masking in `GuidelineWriter` and not find it. Revert that if you disagree with the framing.

Two things deliberately left alone, worth noting so nobody files them as regressions: `MarkdownFormatter` still collapses blank lines in user `.ai/guidelines/*.md` prose outside fences (Boost is assembling that into its own block), and the writer still `rtrim()`s the file before appending its `===` separator.

Didn't add a `FencesTest` — `MarkdownFormatterTest` already covers the masking edge cases that matter (unclosed fence, inline backtick run, literal placeholder in content), and moving them would just be churn.

Full suite green (1058 passed), Pint and PHPStan clean.

PR #1001, opened 2026-09-05
Add zcode agent support

codearachnid:
This is a 2 part PR to add support for nested config keys as common pattern, and registers ZCode agent support.

**1/2 — Nested config keys**

Config keys may use dot notation `mcp.servers` to address nested objects, matching `data_get()`. Keys with a literal dot are passed as an array of segments instead. The JSON and JSON5 writers create any missing segments and preserve sibling keys. Amp's `amp.mcpServers` now uses this; its written output is unchanged for compat.

**2/2 — ZCode agent**

Registers ZCode for boost:install:

- Detection: global `~/.zcode`, or a project `.zcode/` directory or config file
- `AGENTS.md`: writes agent guidelines on install
- Skills: installs to `.zcode/skills`
- MCP: writes servers to `.zcode/config.json`, nested at `mcp.servers`

### DIFF
--- a/src/BoostManager.php
+++ b/src/BoostManager.php
@@ -18,6 +18,7 @@ use Laravel\Boost\Install\Agents\Junie;
 use Laravel\Boost\Install\Agents\Kiro;
 use Laravel\Boost\Install\Agents\OpenCode;
 use Laravel\Boost\Install\Agents\Pi;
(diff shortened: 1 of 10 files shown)


pushpak1300:
Thanks for your pull request!

PR #1005, opened 2026-09-07
Simplify Herd home path assignment

Mohammad-Ranjbar:
Herd::getHomePath() currently uses an explicit isset check before falling back to USERPROFILE on Windows.

Using null coalescing assignment preserves the existing behavior for missing or null HOME values while resolving the IfToNullCoalescingAssignRector finding.


pushpak1300:
Not sure what problem is solving ?

PR #1007, opened 2026-09-07
Fix: browser-logs watcher can crash any Livewire page via JSON.stringify auto-invoking $wire's fake toJSON

### LINKED ISSUE #1006: Browser-logs watcher can crash any Livewire page: JSON.stringify auto-invokes $wire's fake toJSON, firing a real server action
## Bug description

`BrowserLogger`'s console patch (`src/Services/BrowserLogger.php`) serializes every `console.log/info/warn/error/table` argument with:

```js
function safeStringify(obj) {
    const seen = new WeakSet();
    return JSON.stringify(obj, (key, value) => { ... });
}
```

`JSON.stringify(obj, replacer)` invokes `obj.toJSON()` **before** the replacer ever runs, for any object that exposes a callable `toJSON` property — the replacer can't intercept that.

Livewire's `$wire` proxy (`generateWireObject` in `livewire/livewire`'s built JS) returns a callable "remote action invoker" for **any** unrecognized property access, and does not special-case `toJSON` the way Alpine's own merge-proxy does (Alpine explicitly returns a safe function for `toJSON` — see `mergeProxyTrap.get` in the same bundle). So `$wire.toJSON` resolves to a function that, when called, queues and sends a real Livewire request for a server action named `toJSON`.

Result: the moment any code — the app's own, a package's, or Livewire's own Alpine plugin — does `console.log(component)` / `console.warn(..., $wire)` / etc. with an object holding a `$wire` reference while Boost's browser-logs watcher is active, it fires a genuine `POST /livewire/update` calling action `toJSON`, which doesn't exist, and the request 500s with:

```
Livewire\Exceptions\MethodNotFoundException
Method or action [toJSON] does not exist on component [some.component].
```

This reproduces on ordinary Livewire/Volt pages just by mounting the component — no explicit debugging call is needed in application code, since it's Livewire/Alpine internals interacting with `$wire`'s proxy `get` trap.

## Steps to reproduce

1. Fresh Laravel 12 + Livewire 3 (Volt) app, `laravel/boost` installed, `APP_ENV=local` (or `APP_DEBUG=true`) so `BOOST_BROWSER_LOGS_WATCHER` is active.
2. Visit any page rendering a Livewire/Volt full-page component.
3. Observe a `POST /livewire/update` failing with `Method or action [toJSON] does not exist on component [...]`.

## Expected behavior

Boost's browser-logs watcher should never be able to trigger side-effecting calls on the page it's instrumenting, regardless of what objects end up in console arguments.

## Suggested fix

Don't let `JSON.stringify` auto-invoke `toJSON` on arbitrary logged objects. Build a plain-data copy by walking **own enumerable keys** first (which never touches a `toJSON` getter/property), then `JSON.stringify` the plain copy:

```js
function toSafeValue(value, seen) {
    if (value === null || typeof value !== 'object') return value;
    if (value instanceof Error) return { name: value.name, message: value.message, stack: value.stack };
    if (seen.has(value)) return '[Circular]';
    seen.add(value);
    if (Array.isArray(value)) return value.map((item) => toSafeValue(item, seen));
    const plain = {};
    for (const key of Object.keys(value)) {
        try { plain[key] = toSafeValue(value[key], seen); }
        catch (e) { plain[key] = '[Unreadable]'; }
    }
    return plain;
}

function safeStringify(obj) {
    return JSON.stringify(toSafeValue(obj, new WeakSet()));
}
```

I've applied exactly this locally (patching the vendor file) and confirmed it resolves the crash without losing any logging fidelity. Happy to open a PR with this change if useful.

## Environment

- laravel/boost: v1.8.9 in our lockfile; confirmed the vulnerable code is unchanged as of v2.7.1 (latest at time of writing)
- laravel/livewire: 3.x (Volt)
- Laravel 12.45.2, PHP 8.4.24

thealejandro:
## Summary

Fixes #1006.

`BrowserLogger`'s console patch stringifies every logged argument with `JSON.stringify(obj, replacer)`. That native call invokes `obj.toJSON()` **before** the replacer runs, for any object exposing a callable `toJSON` property.

Livewire's `$wire` proxy returns a callable "remote action invoker" for *any* unrecognized property access and does not special-case `toJSON` (Alpine's own merge-proxy does guard it explicitly). So logging anything that references `$wire` while the browser-logs watcher is on fires a real `POST /livewire/update` for a nonexistent `toJSON` action, and the request 500s with `Livewire\Exceptions\MethodNotFoundException`. This reproduces on a stock Livewire/Volt page just by mounting a component — no explicit `console.log` of the component needed in app code.

## Fix

Stop depending on `JSON.stringify`'s automatic `toJSON` invocation. Walk own enumerable keys ourselves into a plain-data copy first (`toSafeValue`), then `JSON.stringify` that copy. Enumerating own keys never touches a `toJSON` getter/property, so this is immune to the class of proxy side effect described above, while producing identical output for ordinary objects/arrays/errors.

## Test plan

- [x] Reproduced the crash on a Laravel 12 + Livewire 3 (Volt) app with `laravel/boost` installed and `BOOST_BROWSER_LOGS_WATCHER` enabled — any page mounting a Livewire component 500s on load.
- [x] Applied this patch locally; the same page now mounts and submits normally, with browser logs still flowing to `/_boost/browser-logs`.
- [ ] No existing automated test covers `BrowserLogger`'s generated script; happy to add a Dusk/browser test if maintainers want one, wasn't sure of the preferred harness for JS-in-Blade in this repo.

### DIFF
--- a/src/Services/BrowserLogger.php
+++ b/src/Services/BrowserLogger.php
@@ -62,23 +62,40 @@ class BrowserLogger
         table: console.table
     };
 
+    // Recursively build a plain-data copy WITHOUT ever letting JSON.stringify's
+    // native toJSON() auto-invocation run on the original value. Some libraries
+    // (e.g. Livewire's `$wire` proxy) return a callable "remote action invoker"
+    // for ANY unrecognized property access, including `toJSON` — so naively
+    // calling JSON.stringify(obj) on such an object fires a real network
+    // request. Walking own-enumerable keys ourselves never touches `.toJSON`.
+    function toSafeValue(value, seen) {
+        if (value === null || typeof value !== 'object') {
+            return value;
+        }
+        if (value instanceof Error) {
+            return { name: value.name, message: value.message, stack: value.stack };
+        }
+        if (seen.has(value)) {
+            return '[Circular]';
+        }
+        seen.add(value);
+        if (Array.isArray(value)) {
+            return value.map((item) => toSafeValue(item, seen));
+        }
+        const plain = {};
+        for (const key of Object.keys(value)) {
+            try {
+                plain[key] = toSafeValue(value[key], seen);
+            } catch (e) {
+                plain[key] = '[Unreadable]';
+            }
+        }
+        return plain;
+    }
+
     // Helper to safely stringify values
     function safeStringify(obj) {
-        const seen = new WeakSet();
-        return JSON.stringify(obj, (key, value) => {
-            if (typeof value === 'object' && value !== null) {
-                if (seen.has(value)) return '[Circular]';
-                seen.add(value);
-            }
-            if (value instanceof Error) {
-                return {
-                    name: value.name,
-                    message: value.message,
-                    stack: value.stack
-                };
-            }
-            return value;
-        });
+        return JSON.stringify(toSafeValue(obj, new WeakSet()));
     }
 
     // Normalize log type for consistency (e.g., 'warn' to 'warning')
--- a/src/Services/BrowserLogger.php
+++ b/src/Services/BrowserLogger.php
@@ -75,6 +75,12 @@ class BrowserLogger
         if (value instanceof Error) {
             return { name: value.name, message: value.message, stack: value.stack };
         }
+        // Date has no own enumerable keys, so the own-keys walk below would
+        // silently collapse it to {}. It's a real (non-Proxy) built-in, so
+        // calling its own native toISOString directly is safe.
+        if (value instanceof Date) {
+            return value.toISOString();
+        }
         if (seen.has(value)) {
             return '[Circular]';
         }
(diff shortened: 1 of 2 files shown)


thealejandro:
Follow-up commit: my own review caught that `Date` instances have no own enumerable keys, so the initial own-keys walk silently collapsed logged `Date` values to `{}`. Added an explicit `instanceof Date` branch that calls `toISOString()` directly — Date is a real built-in (not a Proxy), so this doesn't reintroduce the original risk. Verified with a small Node repro (Date, Error, circular refs, a fake $wire-like proxy, arrays with `undefined`) that output now matches native `JSON.stringify` behavior for all normal values while never invoking the hostile `toJSON`.

PR #1015, opened 2026-09-09
Skip versionless packages in documentation searches

Mohammad-Ranjbar:
The search-docs tool currently appends .x to every package's major version. When Laravel Roster cannot determine a package version, Package::major() returns null, causing Boost to send .x as the package version.

Requests containing this invalid version return no documentation results.

This change excludes packages without a detectable major version from the documentation search payload while preserving all versioned packages.

Regression coverage verifies that a versionless package is omitted and a package with a known major version is still sent correctly.


pushpak1300:
Thanks for your pull request!

PR #1016, opened 2026-09-09
Report failed skill syncs during installation

Mohammad-Ranjbar:
## Summary

- detect failed skill sync results during installation
- report the affected skill names instead of displaying a successful installation
- add a regression test covering multiple failed skills

## Problem

`SkillWriter::sync()` can return `SkillWriter::FAILED` for one or more skills. The install command previously ignored these results and displayed a success indicator for the agent, even though some skills were not installed.

## Solution

Inspect the results returned by `SkillWriter::sync()` and throw an exception containing the failed skill names when any result has a `FAILED` status. The existing per-agent error handling then displays the installation as failed with a useful error message.

## Testing

- `1` new test with `2` assertions
- `121` related tests with `390` assertions
- PHPStan
- Laravel Pint
- `git diff --check`


pushpak1300:
How to reproduce this ? @Mohammad-Ranjbar

Mohammad-Ranjbar:
Thanks for asking. I reproduced this with a package-provided skill containing a non-Markdown asset.

If an existing generated skill directory cannot replace that asset—for example, because both the asset and its containing directory are not writable—the suppressed `copy()` call returns `false`. `SkillWriter::sync()` then returns `SkillWriter::FAILED` for that skill without throwing an exception.

Before this change, `installFeature()` ignored the returned status and displayed a successful `✓` for the agent even though the asset was not updated. With this change, it displays `✗` and reports the affected skill names.

I also replaced the reflection-based test with a command-level regression test. It creates two skills with unwritable generated assets, runs `boost:install --skills --no-interaction`, and verifies that both failed skill names are reported. @pushpak1300

PR #841, opened 2026-06-05
feat: add Pi agent support

cyrodjohn:
## Summary
- add Pi as a supported Boost agent
- detect Pi from the `pi` binary and `.pi/settings.json` project config
- install Pi guidelines to `AGENTS.md` and skills to `.pi/skills`

## Test Plan
- `vendor/bin/pest tests/Unit/BoostManagerTest.php tests/Unit/Install/AgentsDetectorTest.php tests/Feature/Install/Agents/AgentPathResolutionTest.php`
- `vendor/bin/pint --test src/BoostManager.php src/Install/Agents/Pi.php tests/Feature/Install/Agents/AgentPathResolutionTest.php tests/Unit/BoostManagerTest.php tests/Unit/Install/AgentsDetectorTest.php`

PR #765, opened 2026-04-11
feat(agents): add Kiro IDE agent support

oniice:
# Add Kiro IDE support

[Kiro](https://kiro.dev/) is Amazon's AI-powered IDE and is gaining traction in the dev community. This PR adds it as a supported agent alongside Cursor, Claude Code, Copilot, etc.

## Changes

- `src/Install/Agents/Kiro.php` — new agent supporting guidelines (`.kiro/steering/boost.md`), skills (`.kiro/skills/`), and MCP (`.kiro/settings/mcp.json`). Detects via `/Applications/Kiro.app` on macOS and `.kiro/` at the project root.
- `src/BoostManager.php` — registers `kiro` in the default agents map.
- `tests/Unit/Install/AgentsDetectorTest.php` — updated to account for the new agent.

PR #347, opened 2025-11-08
Add Kiro IDE support

jotafurtado:
# Add Kiro IDE Support

## Description
This PR adds support for Kiro IDE to Laravel Boost.

## Changes
- Created `Kiro` CodeEnvironment class implementing both `Agent` and `McpClient` interfaces
- Registered Kiro in `BoostManager`
- Configured MCP installation at `.kiro/settings/mcp.json`
- Configured AI guidelines at `.kiro/steering/laravel-boost.md`
- Added platform detection for macOS, Linux, and Windows
- Updated tests to include Kiro in all relevant test cases
- Updated README with Kiro setup instructions
- Updated CHANGELOG

## Kiro Configuration
When users run `php artisan boost:install`, Kiro will:
- Appear in the list of available editors (for MCP configuration)
- Appear in the list of available agents (for AI guidelines)
- Auto-detect if `.kiro` directory exists in the project
- Create MCP config at `.kiro/settings/mcp.json`
- Create AI guidelines at `.kiro/steering/laravel-boost.md`

PR #1030, opened 2026-09-17
feat: add Oh My Pi (omp) agent support

jotafurtado:
I use omp daily and noticed Boost lacked support for it. This PR adds first-class detection and configuration for omp, so people running it get Boost guidelines, skills, and MCP set up without hand-editing project files.

Implementation follows Pi.php (#841). I added src/Install/Agents/Omp.php and registered 'omp' in BoostManager. Detection checks for the omp binary and ~/.omp on the host, plus .omp/ and .omp/mcp.json in the project. Guidelines go to AGENTS.md, skills to .omp/skills, and MCP to .omp/mcp.json under the mcpServers key, per the upstream mcp-config.md (https://github.com/can1357/oh-my-pi/blob/main/docs/mcp-config.md) and skills.md (https://github.com/can1357/oh-my-pi/blob/main/docs/skills.md).

For context, omp is #6 on OpenRouter's monthly coding ranking with 6.21T tokens, ahead of Codex (https://openrouter.ai/apps/category/coding?period=month). It shares lineage with pi but the binary, config, and skill paths differ, so the pi detector never matched it.

All tests pass, including the new OmpTest, and Pint is clean. Additive only, no existing behavior touched.


pushpak1300:
Thanks for your pull request!

jotafurtado:
This is frustrating to experience for the second time.

Back in November 2025, I opened PR #347 to add support for Kiro IDE. It was closed with this exact canned response citing maintenance concerns, suggesting I publish a separate package instead. I followed that recommendation and built `jcf/boost-for-kiro-ide`. Five months later, PR #765 added Kiro support from another contributor and was merged without any reference to my previous PR. That merged version was actually more basic than what I had already implemented: Boost simply pointed guidelines to `AGENTS.md`, while my package converts MCP prompts into native Kiro steering files and slash commands.

Now I opened PR #1030 to add support for Oh My Pi (omp), and it was closed in under 30 minutes with the same copy-pasted refusal.

Closing working contributions immediately with generic templates, only to accept the same feature later down the road, is discouraging for contributors. If the project genuinely cannot take on new agent integrations, that standard should be applied consistently.

pushpak1300:
@jotafurtado 

There really isn’t a single “right” point at which we decide to support a new Agent. We usually wait until we’re seeing enough adoption and feel comfortable taking on the integration, because once something becomes part of Boost, we effectively have to maintain it indefinitely as that tool evolves.

That also means a contribution we decline today isn’t necessarily something we would never accept. The ecosystem changes, adoption changes, and our willingness to take on that maintenance can change with it.

On the Kiro PR specifically, I agree that we should have connected the later contribution back to your original work. That’s something I could have handled better. At the same time, with the volume of issues and PRs across Boost and the other repositories we maintain, keeping track of every related contribution from months earlier is unfortunately difficult.

These decisions ultimately involve my judgment about what we’re ready to support at a particular point in time. They aren’t based on the contributor, and there’s certainly nothing personal here.

We’ve also intentionally added an extension point in Boost for cases like this. If an integration isn’t something we’re ready to maintain in core yet, it can live in a third-party package, register its own Boost extension, and work normally for people using that tool. That gives integrations room to exist and grow without requiring Boost itself to take on the long-term maintenance immediately.

PR #1032, opened 2026-09-18
Use AGENTS.md for Claude Code guidelines

pushpak1300:
World is healing. 

https://x.com/trq212/status/2101009392611278961


sunchayn:
This assumes that all users have a certain claude version (that supports Agents.md) and that they have the new behavior configured properly to use agents.md

The default is `claude or agents`. So users with claude.md will get that one used instead. With this change, the user with right version and default config will end up with a stale claude.md file if boost doesn't delete the claude.md first during syncing. And users with unsupported version will end up without guidelines.

albertoarena:
> This assumes that all users have a certain claude version (that supports Agents.md) and that they have the new behavior configured properly to use agents.md
> 
> The default is `claude or agents`. So users with claude.md will get that one used instead. With this change, the user with right version and default config will end up with a stale claude.md file if boost doesn't delete the claude.md first during syncing. And users with unsupported version will end up without guidelines.

I agree. We can't break work of users that use an older version of claude. Moving to AGENTS.md is really good, but CLAUDE.md should be still supported.

pushpak1300:
Agree to all of the comments. Claude cli update's automatically and this changes will land in updated version. If someone want to still end up usign the CLAUDE.md they can do it. by setting in config `boost.agents.claude_code.guidelines_path`. 

This will not be deleting exiting `claude.md`

gldrenthe89:
Thanks for the work on this, but I think this change should be reconsidered.

Claude Code 2.1.277 only uses AGENTS.md as a fallback when there is no CLAUDE.md, and CLAUDE.md still takes precedence. Since Boost doesn't delete the existing CLAUDE.md, every existing project will keep loading the old CLAUDE.md while the updated guidelines are written to AGENTS.md and never read. That fails silently, which is the worst kind of failure here.

On top of that, AGENTS.md support depends on feature flags, so it isn't available on Bedrock, Vertex, with telemetry disabled, or on older Claude Code versions. Those users get no guidelines at all on new projects.

The config option works, but it means adding boost.agents.claude_code.guidelines_path to every project, or cleaning up CLAUDE.md by hand, just to keep things working after an update. That feels like a regression.

Would you be open to keeping CLAUDE.md as the default for Claude Code and making AGENTS.md opt-in? That matches how Claude Code itself treats the two files, and teams that want a single shared AGENTS.md can still switch.

mozex:
> Thanks for the work on this, but I think this change should be reconsidered.
> 
> Claude Code 2.1.277 only uses AGENTS.md as a fallback when there is no CLAUDE.md, and CLAUDE.md still takes precedence. Since Boost doesn't delete the existing CLAUDE.md, every existing project will keep loading the old CLAUDE.md while the updated guidelines are written to AGENTS.md and never read. That fails silently, which is the worst kind of failure here.
> 
> On top of that, AGENTS.md support depends on feature flags, so it isn't available on Bedrock, Vertex, with telemetry disabled, or on older Claude Code versions. Those users get no guidelines at all on new projects.
> 
> The config option works, but it means adding boost.agents.claude_code.guidelines_path to every project, or cleaning up CLAUDE.md by hand, just to keep things working after an update. That feels like a regression.
> 
> Would you be open to keeping CLAUDE.md as the default for Claude Code and making AGENTS.md opt-in? That matches how Claude Code itself treats the two files, and teams that want a single shared AGENTS.md can still switch.

I completely agree with this. I think this should be reconsidered.

pushpak1300:
There are a few cases we need to account for:

1. **New applications with the latest Claude Code**
   We should use `AGENTS.md`.

2. **Existing applications with the latest Claude Code**
   We should also use `AGENTS.md`.

3. **New applications with an older Claude Code version**
   These would need `CLAUDE.md`. I think this case is probably less important, since many users may not even have a `CLAUDE.md` yet and Claude Code can update things automatically behind the scenes.

4. **Existing applications with an older Claude Code version**
   These would need `CLAUDE.md`, mainly in cases where the user manually ran a Composer update.

A reasonable middle ground could be to detect whether the repository already has a `CLAUDE.md`. If it does, we continue syncing it as usual.

I’m leaning toward dropping support for case 3. The behaviour there can vary quite a bit, and if instructions are missing in that specific scenario, I think that’s an acceptable tradeoff.

gldrenthe89:
> A reasonable middle ground could be to detect whether the repository already has a `CLAUDE.md`. If it does, we continue syncing it as usual.

Thanks for the quick reply!

I agree that the suggested middle ground, detecting whether the repository already has a CLAUDE.md, is the way to go. That way existing projects won't run into any regressions.

I'm also with you on older Claude Code versions. There are too many variables to account for, and the config option works well as a fallback. Since that only applies to new projects, it's hardly any extra setup work.

Detecting the installed Claude Code version could be another option, but that feels like an edge case to me.

albertoarena:
> Thanks for the work on this, but I think this change should be reconsidered.
> 
> Claude Code 2.1.277 only uses AGENTS.md as a fallback when there is no CLAUDE.md, and CLAUDE.md still takes precedence. Since Boost doesn't delete the existing CLAUDE.md, every existing project will keep loading the old CLAUDE.md while the updated guidelines are written to AGENTS.md and never read. That fails silently, which is the worst kind of failure here.
> 
> On top of that, AGENTS.md support depends on feature flags, so it isn't available on Bedrock, Vertex, with telemetry disabled, or on older Claude Code versions. Those users get no guidelines at all on new projects.
> 
> The config option works, but it means adding boost.agents.claude_code.guidelines_path to every project, or cleaning up CLAUDE.md by hand, just to keep things working after an update. That feels like a regression.
> 
> Would you be open to keeping CLAUDE.md as the default for Claude Code and making AGENTS.md opt-in? That matches how Claude Code itself treats the two files, and teams that want a single shared AGENTS.md can still switch.

I agree

